PR #1, opened 2021-06-08
Vcon 67 add colors

shewood:
[vcon-67](https://decdvirtualconcierge.atlassian.net/jira/software/projects/VCON/boards/2?selectedIssue=VCON-67)

added colors from Figma to tailwind.config.
to use them just preface where you want to use the color like bg-success-main for example. 
There are a bunch of different gray colours so they are combined in one list as the colours are not specific to text or backgrounds but apply to  both. 
The larger grayscale wasn't added at this point but can be if desired by all. 
Added lato and noto fonts.


DaceyTom2 (approved):
Looks good

PR #2, opened 2021-06-08
added messages container

DaceyTom2:
Description

Currently built out the messages container with conversational area hidden for medium screens. The #DEDEDE border colour, handling small screen conversational area and height will need to be accounted for in future PR's

image
image


nibivi77 (approved):
This looks good to me. We can fix the Width once we combine all upcoming components.

PR #4, opened 2021-06-09
Vcon 65 buttons

TishamIslam:
## Description 

Added a responsive button for reporting any feedback.

## How to test the PR

1. Checkout this branch (vcon-65-buttons)
2. npm run dev
3. change screen width to see how it responds

## Things to consider

This button is not in the same container as the conversation window, I'm not 100% sure if that's fine, but it should be next to the footer as it is right now regardless.


src/components/molecules/ButtonFooter.vue
```diff
@@ -0,0 +1,20 @@
+<template>
+    <div class="flex flex-col w-full items-center container mx-auto md:p-5 md:block">
```
nibivi77, line 2: If there is only one button, we don't need this div.

src/components/molecules/ButtonFooter.vue
```diff
@@ -0,0 +1,20 @@
...
+        <long-button colorClass="bg-btn-dark text-gray-light" text="Provide Feedback" />
```
nibivi77, line 3: bg-btn-dark is not the same colour as in Figma. I just merge this branch to main and you can use the colour "primary-dark".

tailwind.config.js
```diff
@@ -40,6 +40,10 @@ module.exports = {
         disabledDK: "#9C9C9C",
         disabledLT: "#D6D6D6",
       },
+      "btn": {
```
nibivi77, line 43: I don't think we need this as you can use the "primary-dark" colour

nibivi77 (changes requested):
I am not sure where the Figma is going but I think we should match it as much as possible.

Here is the difference:
image

i also left a couple comments to help out. :) Let me know if I am wrong about anything !

src/App.vue
```diff
@@ -12,6 +12,7 @@
       <main>
         <router-view/>
       </main>
+      <long-button colorClass="bg-primary-dark text-gray-light" text="Provide Feedback" />
```
nibivi77, line 15: Maybe have the button in the <main> for screen readers. It is part of the main content.

src/components/atoms/LongButton.vue
```diff
@@ -0,0 +1,18 @@
+<template>
```
nibivi77, line 1: Instead of calling this component "LongButton", I think it would be a good idea giving it a more meaningful name like "ProvideFeedbackButton" or something along the lines.

nibivi77 (approved):
This is perfect. I will just leave some comments and its up to you if you want to make the change. Its totally optional :)

src/components/atoms/LongButton.vue
TishamIslam: I didn't see this before I merged the branch, but I think that a feedback section would only make sense to add if there was more than one feedback element.

neilmispelaar: That's fair, but the section wouldn't just have a button though. 

For correct document structure under WCAG you would need something like:

```
<main>
  ...
  <section>
    <h2>Feedback section</h2>
    <button></button>
  </section>
</main>
```

Otherwise without that `<section>` element the button would fall under the last heading in main (in this case the last heading might be the conversation window) and the button doesn't below to that document section.

TishamIslam: That makes sense, but given that it isn't included in the design as of now, it might be best to implement this after a design discussion with Thomas? Though on the other hand it seems a bit too much overhead for just adding a heading and section tag. What do you think?

PR #5, opened 2021-06-09
Vcon 61 message inbox

DaceyTom2:
[VCON-61 Message Inbox](https://decdvirtualconcierge.atlassian.net/browse/VCON-61?atlOrigin=eyJpIjoiZDg0NDA5MWQ4ZDlhNDk1MDg3NzRhMmNlMTcyY2YwMWUiLCJwIjoiaiJ9)

**Description**

Added inbox and inbox items.
There are a combination of inbox items... Unread, Read, Selected, Hover and Focused.
Currently 3 examples are hard coded in the following order Unread, Selected and Read.

You can test Hover and Focus by hovering your cursor over or tabbing to focus.

**Note**
The colors are not 100% accurate and will need to be updated in a future PR
The font style for Nato light text is not applied as it doesn't exist in Google Fonts
The Icon isn't created yet so J is a placeholder
There is no build in logic so inbox items are hard coded

image

### DIFF
--- a/src/components/molecules/Inbox.vue
+++ b/src/components/molecules/Inbox.vue
@@ -1,5 +1,5 @@
 <template>
-  <h1 class="text-4xl">Inbox</h1>
+  <h1 class="font-heading text-4xl p-2">Inbox</h1>
   <inbox-item />
   <inbox-item dayRead="Weds" selected='true' />
   <inbox-item dayRead="Weds" />
--- a/src/components/molecules/Inbox.vue
+++ b/src/components/molecules/Inbox.vue
@@ -1,8 +1,8 @@
 <template>
   <h1 class="font-heading text-4xl p-2">Inbox</h1>
-  <inbox-item />
-  <inbox-item dayRead="Weds" selected='true' />
-  <inbox-item dayRead="Weds" />
+  <inbox-item tabindex="0" />
+  <inbox-item tabindex="0" dayRead="Weds" selected='true' />
+  <inbox-item tabindex="0" dayRead="Weds" />
 </template>
 
 <script lang="ts">
--- a/src/components/molecules/Inbox.vue
+++ b/src/components/molecules/Inbox.vue
@@ -1,0 +1,20 @@
+<template>
+  <h1 class="text-4xl">Inbox</h1>
+  <inbox-item />
+  <inbox-item dayRead="Weds" selected='true' />
+  <inbox-item dayRead="Weds" />
+</template>
+
+<script lang="ts">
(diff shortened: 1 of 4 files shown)


src/components/organisms/MessagesContainer.vue
```diff
@@ -1,22 +1,27 @@
 <template>
   <div class="flex">
-    <div class="w-full md:w-1/4 md:border md:border-r-0 md:border-gray">Inbox</div>
+    <div class="w-full md:w-1/3 md:border md:border-r-0 md:border-gray">
+      <inbox />
```
nibivi77, line 4: We could make the inbox scrollable if there are more inbox items given. (overflow-auto or overflow-y-auto)

src/components/molecules/Inbox.vue
```diff
@@ -0,0 +1,20 @@
+<template>
+  <h1 class="font-heading text-4xl p-2">Inbox</h1>
+  <inbox-item tabindex="0" />
```
nibivi77, line 3: Each InboxItem could have a cursor-pointer to promote the fact that they are clickable.

src/components/molecules/InboxItem.vue
```diff
@@ -0,0 +1,33 @@
+<template>
+  <div :class="[selected ? 'bg-blue-100':'', 'flex items-center w-full h-16 md:h-20 border-2 border-transparent focus:border-black hover:bg-gray-100']">
+    <div class="p-4">
+      <p class="bg-primary-light rounded-full flex items-center justify-center font-heading w-10 h-10 md:w-12 md:h-12 text-3xl">
+        J
```
nibivi77, line 5: I recognize you said that the J is a placeholder. We don't know if that will be an image or a letter that is being passed a prop. 
Regardless, something will be passed down so maybe we can add that as a prop as well? Maybe just the letter for now.

nibivi77 (commented):
This is great! Considering we are not 100% sure how this will interact, I am not requesting a change. Let's talk about  my comments and I will approve it afterwards. :)

src/components/molecules/InboxItem.vue
DaceyTom2: I set it a prop type string. When an icon gets made I hope this will be split out into an atom component instead of roughed in

src/components/molecules/Inbox.vue
DaceyTom2: This is a really good idea. Adds a lot of personality and makes for a more intuitive experience

nibivi77 (approved):
Noice!

PR #7, opened 2021-06-10
removed links and heading

DaceyTom2:
Description

Removed extra links and heading

Removed "This is the home page" heading from home page
Removed Conversation and Notice links from home page

image

### DIFF
--- a/src/router/index.js
+++ b/src/router/index.js
@@ -1,8 +1,6 @@
 import { createRouter, createWebHistory } from 'vue-router';
 import Home from '../views/Home.vue';
 
-const VCConversation = () => import('../views/VC-Conversation.vue');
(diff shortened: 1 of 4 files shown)


TishamIslam (approved):
Looks good, no issues found.

nibivi77 (commented):
We could remove the pages inside the views folder as well to clean it up.

shewood (changes requested):
I am not looking at the whole thing but don't we want to keep the <h1> as per Neil's drawing

DaceyTom2:
@shewood We will want a H1. Currently "Inbox" is a H1 but that can be bumped to H2. As for what the H1 should be is left to be designed so I think it would be fair to add it later

shewood (approved):

PR #10, opened 2021-06-10
Added the option buttons inside Textinput.vue

nibivi77:
### Description

[Prompt cards (yes/no)](https://decdvirtualconcierge.atlassian.net/jira/software/projects/VCON/boards/2?selectedIssue=VCON-54)
Added the optional buttons to answer the VC. 
It displays dynamically according to how many options are passed as props.

### Note
Since we do not know exactly how the options are provided, the props are passed as simple strings in an array.

The design on the card is wrong. The correct one is on Figma:
image


DaceyTom2 (approved):
Looks good, we will have to circle back once the color changes are merged to main as Sheila already did a passby

shewood (approved):
looks nice

PR #11, opened 2021-06-10
Added images.

nibivi77:
### Description

Added images from the Figma designs.

Not much else to say...
This PR seems too small so I will add a little bit of text to make it seem a little bit bigger than if it was empty. i hope you don't mind me wasting yo time reading this.


DaceyTom2 (approved):
Lookin good, would be cool to see this icon as the favicon but I imagine all GOC uses the red leaf

PR #12, opened 2021-06-10
Vcon 52 banner

shewood:
[vcon-52]

Added the banner to the header.


src/components/atoms/Banner.vue
```diff
@@ -0,0 +1,51 @@
...
+          uppercase
+          bg-blue-deep bg-opacity-95
+          text-white
+          inline-block
+          whitespace-normal
+        "
+      >
+        {{ title }}
+      </h1>
+      <div class="">
```
nibivi77, line 20: Can we remove this div and add CSS to the p tag instead?

src/components/molecules/TheHeader.vue
```diff
@@ -120,16 +121,21 @@
           text="menu"
         ></Menu>
       </div>
+      <div class="sm:py-4">
```
nibivi77, line 124: Can we discard this div and apply the sm:py on the wrapper div inside the "TheHeader" component?

nibivi77 (commented):
This looks good following the Figma. The mobile does not look the same but I assume the banner is subject to change.

Let me know what you think about the comments and I will approve.

src/components/atoms/Banner.vue
```diff
@@ -0,0 +1,51 @@
+<template>
+  <div class="bg-banner-img bg-cover p-10">
+    <div class="mx-auto container px-3">
+      <h1
```
DaceyTom2, line 4: The h1 and p tag overlap slightly leaving a dark line
image

DaceyTom2 (changes requested):

src/components/atoms/Banner.vue
shewood: i took out the empty class but can't get rid of the div.

src/components/atoms/Banner.vue
shewood: yes, it is overlapping.

DaceyTom2 (approved):
Looks beautiful!

PR #13, opened 2021-06-11
Added prettier on commit with husky

DaceyTom2:
Description

Added prettier to lint code
Husky will lint with prettier on pre commit

Note
Vue files, App, model and index were changed to comply with prettier


nibivi77 (approved):
Thank you for this!

src/components/molecules/InboxItem.vue
```diff
@@ -1,34 +1,69 @@
...
+          text-3xl
+        "
+      >
+        {{ icon }}
       </p>
     </div>
     <div class="flex-auto truncate">
-        <p :class="[!dayRead ? 'font-bold':'', 'font-body md:text-lg truncate overflow-ellipsis text-gray-dark']">Virtual Concierge</p>
-        <p class="font-body font-light text-sm md:text-base truncate overflow-ellipsis text-gray-dark">Hi Mary, I'm your Virtual Con...</p>
+      <p
```
shewood, line 27: I find this hard to read, can you make it so that it doesn't break up the expression so much?

shewood (changes requested):

src/components/molecules/InboxItem.vue
DaceyTom2: I'd have to disagree, I find this much cleaner than the single line. These are prettier's default rules but if the teams wants we can come up with a formatting that fits all of our needs better

shewood: I was thinking more like how it was done in the benefits finder,

shewood (approved):
I don't like it but I can live with the changes that prettier makes.

PR #14, opened 2021-06-11
Vcon 59 messages(job alert)

TishamIslam:
## Description
This holds something similar to email messages for the VC, which (I assume) can come from several sources, with the given example of the Job Bank. The list is scrollable, and for now the messages have a bunch of dummy text. There's no prop logic built-in yet for the larger components, but I commented out my attempt at implementing it. 
Also, there needs to be a list of paragraphs for the MessageCard, which I imagine would be implemented with an array and v-for, but I'd need to look into that a bit more + right now we're not implementing that kind of logic.

The component hierarchy is as follows: MailContainer holds the MessageHeader (icon + sender) and the MessageList (has all the messages). The MessageList contains multiple MessageCards

Additionally, I refactored the ConversationWindow's header to be the same MessageHeader I am using, as they are the same thing but with different icons + header text. 

## Notes
I commented out Ben's conversation window so that this could display, as we'd need some additional logic to switch from this window to Ben's conversation window. 
I uploaded the image files separately myself, so I'm going to need merge with main and then refactor my code so that the icons display properly after the merge.


src/components/organisms/ConversationWindow.vue
```diff
@@ -1,33 +1,25 @@
 <template>
-    <div class="w-full h-full flex flex-col">
-        <div class="flex border-b border-gray-infomd p-2">
-            <img src="../../assets/UserAccount/VirtualConcierge.svg" alt="Bot Image." class="h-10 mt-auto w-10">
```
DaceyTom2, line 4: Some of this didn't merge well. The bot logo is missing and the alignment is off

src/components/organisms/MailContainer.vue
```diff
@@ -0,0 +1,32 @@
+<template>
+  <div class="flex">
```
DaceyTom2, line 2: We can reuse message container for this. instead of swapping out the containers swap out the ConversationWindow component

DaceyTom2 (changes requested):

src/components/organisms/MailContainer.vue
TishamIslam: Benoit and I talked about it and we couldn't immediately find a way to pass in a child component, so we decided to implement separate components instead for the time being.

DaceyTom2: We can do what you are currently doing by commenting out the component but inside of the container

shewood (changes requested):
can you add font-heading to the title and font-body for the content.

PR #17, opened 2021-06-11
added custom height of 50% viewport

DaceyTom2:
Description

Can now set custom heights in tailwind.
Messages container is leveraging the 1/2 viewport height

### DIFF
--- a/src/components/organisms/MessagesContainer.vue
+++ b/src/components/organisms/MessagesContainer.vue
@@ -1,5 +1,5 @@
 <template>
-  <div class="flex md:h-96">
+  <div class="flex md:h-vh-1/2">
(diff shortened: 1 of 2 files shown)


nibivi77 (approved):
Yes, yes, yes good. Its as good as a purple drift in Mario Kart 8 deluxe.

TishamIslam (changes requested):
I think that it needs to be bigger, the conversation window looks quite short on my screen, maybe something like 75vh would work better?

shewood (approved):
Looks good

neilmispelaar:
@ThomasFHoy Any opinions on height size of the chat window / inbox

TishamIslam (approved):

PR #19, opened 2021-06-11
Banner change

shewood:
Updated the banner for small screens so that the text and blue area fill more of the screen, IT should look the same on larger screens.


DaceyTom2 (changes requested):
Smaller screen sizes the text is getting cut off

DaceyTom2 (approved):
I'm silly, your work is clean!

PR #20, opened 2021-06-11
Vcon 59 messages

TishamIslam:
## Description
This branch is more accessible because the name doesn't have \() 
This holds something similar to email messages for the VC, which (I assume) can come from several sources, with the given example of the Job Bank. The list is scrollable, and for now the messages have a bunch of dummy text. There's no prop logic built-in yet for the larger components, but I commented out my attempt at implementing it. 
Also, there needs to be a list of paragraphs for the MessageCard, which I imagine would be implemented with an array and v-for, but I'd need to look into that a bit more + right now we're not implementing that kind of logic.

The component hierarchy is as follows: MailContainer holds the MessageHeader (icon + sender) and the MessageList (has all the messages). The MessageList contains multiple MessageCards

Additionally, I refactored the ConversationWindow's header to be the same MessageHeader I am using, as they are the same thing but with different icons + header text. 

## Notes
I commented out Ben's conversation window so that this could display, as we'd need some additional logic to switch from this window to Ben's conversation window. 
I uploaded the image files separately myself, so I'm going to need merge with main and then refactor my code so that the icons display properly after the merge.


src/views/Home.vue
```diff
@@ -1,17 +1,20 @@
 <template>
   <div class="home">
-    <messages-container />
+    <!-- <messages-container /> -->
+    <mail-container />
```
nibivi77, line 4: The mail-container should be inside the messages-container, right next to the ConversationMessage component. 

Doing this will avoid using two different inbox components. We do not want 2 different inbox components as they are the same for both the job banks(your component) and the Virtual Concierge(ConversationMessage component).

src/components/atoms/MessageHeader.vue
```diff
@@ -0,0 +1,20 @@
...
+    <div class="flex border-b border-gray-200 p-2">
+        <img :src="'../src/assets/' + imagePath + imageName" :alt="altText" class="h-10 mt-auto w-10">
+        <span class=" font-bold pt-2 pl-2">{{ headerText }}</span>
+    </div> 
+</template>
+
+<script>
+export default {
+    props: {
+        imagePath: String,
```
nibivi77, line 11: Do we need the image path prop? We can combine the image path and image name into one prop.

src/components/molecules/MessageCard.vue
```diff
@@ -0,0 +1,33 @@
+<template>
+    <div class="bg-white p-6 space-y-3" >
+        <span class="uppercase text-sm font-light font-body">{{ timestamp }}</span>
+        <h1 class="font-title text-2xl font-semibold">{{ title }}</h1>
+        
+        <!-- <p class="font-normal text-lg" v:for="paragraph of paragraphs">
+            {{ paragraph }}
+        </p> -->
```
nibivi77, line 8: Clean up the comments if they are not needed :) Or leave a comment explaining what they are used for.

src/components/molecules/MessageCard.vue
```diff
@@ -0,0 +1,33 @@
...
+            {{ paragraphs }}
+        </p>
+    </div>
+    <!-- <div class="bg-white p-6 space-y-3" >
+        <span class="uppercase text-sm font-light">Weds, May 14th, 10:04 am</span>
+        <h1 class="text-2xl font-semibold">Your Employment Insurance Claim - You've been subscribed to Job Alerts</h1>
+        <p class="font-normal text-lg"> 
+            Lorem ipsum dolor sit amet consectetur adipisicing elit. Harum ea eos ducimus nulla sit quae, commodi eaque ut eveniet laudantium amet maxime consectetur, facere repellendus ipsum tenetur numquam officia mollitia?
+        </p>
+    </div> -->
```
nibivi77, line 19: Clean up the comments if they are not needed :) Or leave a comment explaining what they are used for.

src/components/molecules/MessageList.vue
```diff
@@ -0,0 +1,46 @@
...
+            title="Your Employment Insurance Claim - You've been subscribed to Job Alerts" 
+            paragraphs="Lorem ipsum dolor sit amet consectetur adipisicing elit. Harum ea eos ducimus nulla sit quae, commodi eaque ut eveniet laudantium amet maxime consectetur, facere repellendus ipsum tenetur numquam officia mollitia?"    
+         />
+         <message-card 
+            timestamp="Weds, May 14th, 10:04 am" 
+            title="Your Employment Insurance Claim - You've been subscribed to Job Alerts" 
+            paragraphs="Lorem ipsum dolor sit amet consectetur adipisicing elit. Harum ea eos ducimus nulla sit quae, commodi eaque ut eveniet laudantium amet maxime consectetur, facere repellendus ipsum tenetur numquam officia mollitia?"    
+         />
+        <message-card />
+        <message-card title="Empty box above" />
```
nibivi77, line 19: line 19 and 18, Not sure if we need those? 

Also, I think it would be good to follow the Figma and only have 1 message card. (You can leave a comment in the page stating that multiple cards can be added to the list. Perhaps leave a message card commented.)

src/components/molecules/MessageList.vue
```diff
@@ -0,0 +1,46 @@
...
+    </div>
+<!--
+<div class="flex flex-col h-96 bg-gray-light p-6 space-y-6 overflow-auto"> 
+    <div class="bg-white p-6 space-y-3" v:for="msg of messages">
+        <span class="uppercase text-sm font-light">{{ msg.timestamp }}</span>
+        <h1 class="text-2xl font-semibold">{{ msg.title }}</h1>
+        <p class="font-normal text-lg">{{ msg.paragraphs }}</p>
+    </div>
```
nibivi77, line 27: Clean up the comments if they are not needed :) Or leave a comment explaining what they are used for.

src/components/molecules/MessageList.vue
```diff
@@ -0,0 +1,46 @@
...
+</template>
+
+<script>
+import MessageCard from "./MessageCard.vue"
+
+export default {
+    //  not 100% sure how to implemnet prop passing
+    //  props: {
+    //     messages: Array, //should be an array of objects with a timestamp, a header, and an array of paragraphs  
+    // },
```
nibivi77, line 38: I agree with passing in an array of messages(objects) with the timestamp, header and text. We can call together and we can see how that is done. :)

i don't think the paragraphs need to be in an array. I believe when we create the data, we can simply add line breaks to the texts. We will see when that is done :p

src/components/organisms/ConversationWindow.vue
```diff
@@ -1,71 +1,25 @@
...
+            <ConversationMessage :isUser=false :text="'I appreciate your concern. Apply water vigorously and call 911.'" />                
+            <ConversationMessage :isUser=true :text="'I need help! My desk is on fire!'" />
+            <ConversationMessage :isUser=false :text="'Hi, how can I help you?'" />
+            <!-- duplicate for content placeholder -->
+            <ConversationMessage :isUser=false :text="'Oh... Well, this is akward.'" />
+            <ConversationMessage :isUser=true :text="'I meant I am working too hard; I am all fired up right now.'"/>
+            <ConversationMessage :isUser=false :text="'I appreciate your concern. Apply water vigorously and call 911.'" />                
+            <ConversationMessage :isUser=true :text="'I need help! My desk is on fire!'" />
+            <ConversationMessage :isUser=false :text="'Hi, how can I help you?'" />
+            <!-- here -->
```
nibivi77, line 17: I think I added this, but could you remove it in your PR please? :) The "here" comment

nibivi77 (changes requested):
This is starting to look good Tisham. A couple of things to review however.

Ask me for any help ! :)

src/components/atoms/MessageHeader.vue
TishamIslam: Oops that's my bad, I didn't want a long alt-text name so I separated the two. Will fix.

src/components/molecules/MessageList.vue
TishamIslam: For now I've added css classes so that line breaks are preserved in the passed prop, but we can change it in the future if need be.

nibivi77 (changes requested):
Much Better! One more thing, I am not sure if this was introduced in this PR but when swapping to the Virtual Concierge chat, it is not aligned anymore. In addition, No images show up.

image

nibivi77 (approved):
Good work ! It would be nice to have some input from someone else on this as well. :)

P.S: If you want, we would need to fix the sizing of the MessageHeader.vue.  From the Figma, it is a little bit more spaced out compared to what we have.

DaceyTom2:
I provided feedback on the previous PR but I think everything was covered. Awesome job Tisham!

DaceyTom2 (approved):

PR #21, opened 2021-06-14
changed button margin

TishamIslam:
Just changed the button margins on small screens, and also fixed the size.


nibivi77 (commented):
I think this PR can be removed. It is also included in your other bigger PR with some feedback already given.

PR #22, opened 2021-06-14
Text input feedback.

nibivi77:
### Description
Quick CSS fixes for the text-input.

- yes/no button background image to white.
- yes/no button hover background changed to gray-infolt
- Send image button is bigger.
- Removed the 1px top white line above the image button.


shewood (commented):
What is supposed to happen to the buttons on the hover? the background looks odd being the same colour as the bar behind.

shewood (approved):
The buttons look good. Follow up on the hover state and update accordingly.

PR #26, opened 2021-06-15
Moved inbox heading down and to the right

DaceyTom2:
Description

Lowered inbox heading down and to the right. It is also now Bold to follow the design.

This has a small impact on conversationWindow that may or may not need addressing


nibivi77 (approved):
Good!

PR #27, opened 2021-06-15
cleaned up inbox items outside of item div

DaceyTom2:
Description

Focus looked off since some inbox items had elements larger than div they were inside.

Cleaned up padding on icon
Cleaned up size in small screens of ReadNotification

### DIFF
--- a/src/components/molecules/InboxItem.vue
+++ b/src/components/molecules/InboxItem.vue
@@ -5,7 +5,7 @@
       'flex items-center w-full h-16 md:h-20 border-2 border-transparent focus:border-black hover:bg-gray-100 cursor-pointer',
     ]"
   >
-    <div class="p-4">
+    <div class="p-2 md:p-4">
       <p
         class="
           bg-blue-light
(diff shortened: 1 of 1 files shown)


src/components/molecules/InboxItem.vue
```diff
@@ -5,7 +5,7 @@
       'flex items-center w-full h-16 md:h-20 border-2 border-transparent focus:border-black hover:bg-gray-100 cursor-pointer',
     ]"
   >
-    <div class="p-4">
+    <div class="p-2 md:p-4">
```
nibivi77, line 8: We can get rid of this div and apply the styling on the p tag. instead of padding it will become margin instead. so: m-2 md:m-4

nibivi77 (changes requested):
Looks good let me know what you think about my comment.

src/components/molecules/InboxItem.vue
DaceyTom2: It's a good suggestion but the `<p>` tag isn't responsive enough. Regardless this p tag is being switched out for an icon soon

nibivi77: ok sounds good

nibivi77 (approved):

PR #29, opened 2021-06-16
Vuex modules setup

nibivi77:
### Description
Added the modules discussed in today's meeting.
Also converted the "Send message" to the VC to use the correct module.

[Vuex Modules](https://decdvirtualconcierge.atlassian.net/jira/software/projects/VCON/boards/2?selectedIssue=VCON-75)

## Notes

- No Tests implemented.
- This is the skeleton and no logic has been added except for the chatMessages to send messages.

## How to Test

1. `git checkout Vuex-modules-setup`
2. Under folder src -> store -> modules.


TishamIslam (commented):
It seems to be working as expected, great work! Given that it's a pretty big PR, I'll let Tom/Sheila approve it.

src/store/modules/inbox.js
```diff
@@ -2,82 +2,23 @@
...
-        {isUser: false, text: 'I appreciate your concern. Apply water vigorously and call 911.'},
-        {isUser: true, text: 'I need help! My desk is on fire!'},
-        {isUser: false, text: 'Hi, how can I help you?'},
-        {isUser: false, text: 'Oh... Well, this is akward.'},
-        {isUser: true, text: 'I meant I am working too hard; I am all fired up right now.'},
-        {isUser: false, text: 'I appreciate your concern. Apply water vigorously and call 911.'},
-        {isUser: true, text: 'I need help! My desk is on fire!'},
-        {isUser: false, text: 'Hi, how can I help you?'}]
-}
+  selectedInboxItemId: 0,
```
DaceyTom2, line 5: By default we will want to have this set to VC chatbot but in a future PR

src/store/modules/chatMessages.js
```diff
@@ -0,0 +1,146 @@
...
+      },
+    ],
+  },
+];
+
+// getters
+const getters = {
+  // Get virtual concierge items
+  getChatMessage(state) {
+    return state[0];
```
DaceyTom2, line 78: We want to get messages in order of datetime

DaceyTom2 (changes requested):
Overall looks really well done. I'd recommend creating a separate PR for formatting issues in the future as this PR is way too large to be easily reviewed and doesn't follow it's own description

src/components/atoms/TextInput.vue
```diff
@@ -29,26 +26,26 @@ export default {
...
-  }, 
+  },
   emits: ["add-message"],
-  setup(_, context){
+  setup(_, context) {
     const text = "";
-    function sendText(){
-      if(this.text.length > 0){
-        context.emit("add-message", this.text)
+    function sendText() {
```
DaceyTom2, line 33: Can this clear the text field as well?

DaceyTom2 (commented):
I was wondering if this can clear the text field as well because clicking multiple times sends the same message

src/store/modules/chatMessages.js
```diff
@@ -0,0 +1,146 @@
...
+      Math.max(...conversation.messages.map((msg) => msg.chatMessageId)) + 1;
+    // Create the new message object.
+    const newMessage = {
+      chatMessageId: nextMessageId,
+      receivedTime: Date.now(),
+      isUser: payload.isUser,
+      text: payload.text,
+    };
+    // Find index of the chat in the state.
+    const index = state.findIndex((chat) => chat.chatId === payload.chatId);
```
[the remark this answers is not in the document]
nibivi77: The index and the conversation.chatId is not the same. Thats why I needed to get the index of where the conversation is in the state (array).

src/store/modules/chatMessages.js
```diff
@@ -0,0 +1,146 @@
...
+    // Add the new message to the correct conversation.
+    state[index].messages.unshift(newMessage); // need to find the index
```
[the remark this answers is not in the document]
nibivi77: When displaying the messages in the VC, they are in reverse. That's why I used unshift instead of push. Maybe we can find a way to work around it?

src/store/modules/inbox.js
nibivi77: Correct, that will be part of whoever is doing the inbox functionality PR.

src/store/modules/chatMessages.js
nibivi77: This Get is to fetch the Conversation object. Not just the messages.
The Get will completely change once we have the inbox "selected item" functionality. 

Right now I hardcoded the first conversation [0] in the state. 

Also, getting the messages by date might not be needed as they are added to the array of messages in chronological order. Ordering can also be done in the front end. Remains to be seen..

DaceyTom2: Ordering by index isn't a good idea. There is no guarantee it is in order, it could potentially be a alphanumeric guid. I'd recommend two getters. One that returns state[0] and one that returns state[0].messages in order

src/store/modules/chatMessages.js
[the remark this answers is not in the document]
DaceyTom2: const conversation = state.find((chat) => chat.chatId === payload.chatId);
and
const index = state.findIndex((chat) => chat.chatId === payload.chatId);
state[index]

Gets the same object but with more steps

src/store/modules/chatMessages.js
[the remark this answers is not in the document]
DaceyTom2: The order in state shouldn't be relevant. It should be done by a getter or within the component

src/store/modules/chatMessages.js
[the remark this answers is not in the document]
nibivi77: One is find() and the other is findIndex. find() returns the object itself, findIndex() returns the index where it is located in the array.

It is not the same. Maybe there is something I am missing?

src/components/atoms/TextInput.vue
DaceyTom2: Also also, if we hit return can it click the send button?

src/store/modules/chatMessages.js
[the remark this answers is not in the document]
DaceyTom2: You are using the index to find an object you already have

src/store/modules/chatMessages.js
nibivi77: I created the two methods in the getter requested.

src/store/modules/chatMessages.js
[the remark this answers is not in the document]
nibivi77: Created a sort array in the getter to enable push instead of unshift.

src/components/atoms/TextInput.vue
nibivi77: I can't get the text to clear but I added the keyup.enter event to send messages.

DaceyTom2 (approved):
Awesome work!

neilmispelaar:
One thing that might make things easier for the text input: 

`const text = "";` 

You could make this a ref (reference to a watched variable) in the setup function and then bind the data to the input field using `v-model`

```
<!-- Text Input HTML - You already got this --> 
<input v-model="text" type="text" class="..." />
...

import { ref } from 'vue'; 

setup() { 
  // Initialise reactive value to a string 
  const text = ref("");
  
  function sendBtnClicked() {
      // dispatch the vuex store action or whatever 

      // Note: we access the value of the ref element using the VALUE property
      console.log(**text.value**);
      
      // Clear the text after the button is clicked 
      **text.value** = "";
      
    }
  
  return { 
    text,
    sendBtnClicked,
  }
}
```

nibivi77:
> One thing that might make things easier for the text input:
> 
> `const text = "";`
> 
> You could make this a ref (reference to a watched variable) in the setup function and then bind the data to the input field using `v-model`
> 
> ```
>  
> <input v-model="text" type="text" class="..." />
> ...
> 
> import { ref } from 'vue'; 
> 
> setup() { 
>   // Initialise reactive value to a string 
>   const text = ref("");
>   
>   function sendBtnClicked() {
>       // dispatch the vuex store action or whatever 
> 
>       // Note: we access the value of the ref element using the VALUE property
>       console.log(text.value);
>       
>       // Clear the text after the button is clicked 
>       text.value = "";
>       
>     }
>   
>   return { 
>     text,
>     sendBtnClicked,
>   }
> }
> ```

Thank you Neil! It worked ! :)

PR #31, opened 2021-06-17
Vcon 73 inbox vuex implementation

DaceyTom2:
Description

Inbox vuex state created

Notes
Code isn't implemented into vue application yet (separate PR)
Still a need to pull information from conversation and email state in separate PR when work is completed


src/store/modules/inbox.js
```diff
@@ -3,17 +3,53 @@
...
+    return inboxItems;
+  },
+};
 
 // actions
-const actions = {};
+const actions = {
+  async selectInboxItem(context, id) {
+    context.commit("updateSelectedInboxItemId", id);
+  },
```
nibivi77, line 44: Do you need to make this async? I don't think there will be an API call here. What do you think?

DaceyTom2: It doesn't but I rather leave it for consistency sake

nibivi77 (approved):
Good! We might not need to have that action method but things are surely going to change as we build this.

src/store/modules/inbox.js
TishamIslam: I think it makes more sense to remove it, since having "async" in front of it makes it _seem_ like it's an async operation but if it isn't expected to be, then it's misleading.

PR #33, opened 2021-06-17
wired inbox vuex data into components

DaceyTom2:
Description

Added functionality to Inbox using the vuex state

To test you can load the page then click on inbox items
Currently implementing changing between types of chat and email

### DIFF
--- a/src/components/molecules/Inbox.vue
+++ b/src/components/molecules/Inbox.vue
@@ -2,8 +2,8 @@
   <h1 class="font-heading font-bold text-4xl pl-3 pt-4">Inbox</h1>
   <div class="md:overflow-auto space-y-1">
     <inbox-item
-      v-for="(inboxItem, index) in inboxItems"
-      :key="index"
+      v-for="inboxItem in inboxItems"
+      :key="inboxItem.id"
       tabindex="0"
       :inboxItem="inboxItem"
       @select-inbox-item="selectInboxItem"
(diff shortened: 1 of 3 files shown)


src/components/molecules/Inbox.vue
```diff
@@ -2,27 +2,30 @@
   <h1 class="font-heading font-bold text-4xl pl-3 pt-4">Inbox</h1>
   <div class="md:overflow-auto space-y-1">
     <inbox-item
+      v-for="(inboxItema, index) in inboxItems"
```
shewood, line 5: is it supposed to be inboxItema?

DaceyTom2: Typo! I fixed it

shewood (approved):
Looks good.

src/components/molecules/Inbox.vue
```diff
@@ -2,27 +2,30 @@
   <h1 class="font-heading font-bold text-4xl pl-3 pt-4">Inbox</h1>
   <div class="md:overflow-auto space-y-1">
     <inbox-item
+      v-for="(inboxItem, index) in inboxItems"
+      :key="index"
```
nibivi77, line 6: Instead of the index, we could use the inboxItem.id for the key. What do you think?

P.S My comment might be completely useless x)

DaceyTom2: It makes more sense to me. I updated it but idk if it will ever get used

nibivi77 (approved):

PR #35, opened 2021-06-17
Vcon 91 email vuex implementation

TishamIslam:
## Description
Added vuex state management to the emails module, along with the appropriate getters/actions/mutators.

## Notes
Lacks unit tests
Needs some kind of date formatting, right now I'm using basic Date methods so that the timestamp is human-readable, but it's not 100% accurate to the design. For that purpose, I used "new Date()" instead of "Date.now()" for the timestamps
I added some extra unused methods that will probably be used at some point, just not now e.g. a method that "selects" one sender for easy accessibility

### DIFF
--- a/src/components/molecules/MessageList.vue
+++ b/src/components/molecules/MessageList.vue
@@ -12,15 +12,6 @@
       />
     </div>
   </div>
-
-  <!-- How the messages might be implemented later: an array prop with message objects iterated over to create a list of MessageCards
-<div class="flex flex-col h-96 bg-gray-light p-6 space-y-6 overflow-auto"> 
-    <div class="bg-white p-6 space-y-3" v:for="msg of messages">
-        <span class="uppercase text-sm font-light">{{ msg.timestamp }}</span>
-        <h1 class="text-2xl font-semibold">{{ msg.title }}</h1>
-        <p class="font-normal text-lg">{{ msg.paragraphs }}</p>
-    </div>
-</div> -->
 </template>
 
 <script>
@@ -30,10 +21,6 @@ import { useStore } from "vuex";
 import { computed } from "vue";
 
 export default {
-  //  not 100% sure how to implemnet prop passing
-  //  props: {
-  //     messages: Array, //should be an array of objects with a timestamp, a header, and a paragraph
-  // },
   components: {
     MessageCard,
     MessageHeader,
--- a/src/components/molecules/MessageList.vue
+++ b/src/components/molecules/MessageList.vue
@@ -21,6 +21,7 @@
 <script>
 import MessageCard from "./MessageCard.vue";
 import MessageHeader from "../atoms/MessageHeader.vue";
+import EmailService from "../../services/emails/emails.js";
 import { useStore } from "vuex";
 import { computed } from "vue";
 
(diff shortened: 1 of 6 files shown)


src/store/modules/emails.js
```diff
@@ -15,13 +26,70 @@ const state = [
...
+    const newEmail = {
+      emailId: emailId,
+      messageTitle: payload.messageTitle,
+      messageBody: payload.messageBody,
+      receivedTime: Date.now(),
+    };
+
+    sender.push(newEmail);
+  },
+  moveSenderToFirst(state, selectedSenderId) {
```
DaceyTom2, line 81: Using a getter is an easier way to select and item instead of moving sender reordering the state.
This can be done by searching for an email id based off of Inbox/selectedInboxItemId in your getEmailsMostRecentFirst getter. If you have questions please ask

DaceyTom2 (changes requested):

src/store/modules/emails.js
```diff
@@ -1,27 +1,95 @@
...
-    emails: [
+const state = {
+  selectedSenderId: 1,
+};
+
+// getters
+const getters = {
+  getEmails() {
+    //this would use the email service
+    let mail = [
```
DaceyTom2, line 9: This object should be retrieved from state. Check out how chatMessages handles getChatMessageObject()

DaceyTom2 (changes requested):

src/store/modules/emails.js
```diff
@@ -1,27 +1,95 @@
...
-const getters = {};
+    return sortedMail;
+  },
+};
 
 // actions
-const actions = {};
+const actions = {
+  //these functions would probably be used by the email service
+  sendEmail({ commit }, payload) {
```
nibivi77, line 64: I do not know if we will ever be able to "send emails" or "add emails". This logic might be useless for a while. (correct me if i am wrong please.)

src/store/modules/emails.js
```diff
@@ -1,27 +1,95 @@
...
+  },
+  senderIsSelected({ commit }, selectedSenderId) {
+    commit("moveSenderToFirst", selectedSenderId);
+  },
+};
 
 // mutations
-const mutations = {};
+const mutations = {
+  addEmail(state, payload) {
```
nibivi77, line 75: Same goes for this functtion. I don't know if we need it.

(We can just leave all this logic here and remove it later)

nibivi77 (commented):
Seems like Tom covered your PR well. Just a comment:

src/store/modules/emails.js
TishamIslam: That is true, I originally thought that there might be a mechanism on the client-side to add emails but obviously that'd be handled by the email service. Given that it's a very simple implementation, I'll remove it and if we ever need to, it won't be a big issue to re-add it.

DaceyTom2 (approved):
Looks good!

src/components/molecules/MessageList.vue
```diff
@@ -21,6 +21,7 @@
 <script>
 import MessageCard from "./MessageCard.vue";
 import MessageHeader from "../atoms/MessageHeader.vue";
+import EmailService from "../../services/emails/emails.js";
```
nibivi77, line 24: The Service should only be imported inside Vuex (our modules).

nibivi77 (approved):
This is good. Check out my comment before merging. I think you just need to remove 1 line of code.

src/components/molecules/MessageList.vue
TishamIslam: Oops, I had put it in there for testing haha

PR #36, opened 2021-06-18
Vcon 92 chat message service

nibivi77:
### Description
Added the [ChatService](https://decdvirtualconcierge.atlassian.net/jira/software/projects/VCON/boards/2?selectedIssue=VCON-92) with a fake JSON file in our public folder, under data.

## Notes

- We need to figure how to handle the dates. for now they are stored in english instead of time. This will affect our sorting logic.
- I am not sure if this PR is using the best practices so please be vigilant when reviewing this one.

### DIFF
--- a/src/store/modules/chatMessages.js
+++ b/src/store/modules/chatMessages.js
@@ -1,87 +1,22 @@
 import ChatMessageService from "../../services/chatMessages/chatMessages";
 
-const state = [];
-// const state = [
-//   {
-//     chatId: 1,
-//     senderName: "Virtual Concierge",
-//     senderIcon: "logo.png",
-//     messages: [
-//       {
-//         chatMessageId: 1,
-//         receivedTime: new Date("August 11, 1975 23:15:30"),
-//         isUser: false,
-//         text: "Hi, how can I help you?",
-//       },
-//       {
-//         chatMessageId: 2,
-//         receivedTime: new Date("August 13, 1975 23:15:30"),
-//         isUser: true,
-//         text: "I need help! My desk is on fire!",
-//       },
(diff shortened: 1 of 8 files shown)


src/store/modules/chatMessages.js
```diff
@@ -1,123 +1,49 @@
...
-        return -1;
-      }
-      if (a.receivedTime < b.receivedTime) {
-        return 1;
-      }
-      return 0;
-    });
-    return conversation;
+    if (state.chatMessages.length !== 0) {
+      const conversation = [...state.chatMessages[0].messages].sort((a, b) => {
```
TishamIslam, line 13: You could use state.chatMessages[0].messages here, since messages is an array.

DaceyTom2: To add onto this we can specify which chatMessages to get now from inbox/selectedInboxItemId. We may need to coordinate this change

DaceyTom2 (approved):
Looks good and keep the change you filthy animal

src/store/modules/chatMessages.js
nibivi77: Taking a look now.

nibivi77: I actually can't do what is suggested because that would change the state. A getter cannot mutate the state.

TishamIslam: Ohh, that makes sense. In that case, could you add a comment to clarify that ? It's not very obvious at first glance that this is a quick way to get a copy of the array.

nibivi77: Sure!

PR #38, opened 2021-06-18
set the height at small screen for the messages

shewood:
set the height to 60 for small screens to hopefully make it so you can see the message window.

### DIFF
--- a/src/components/organisms/MessagesContainer.vue
+++ b/src/components/organisms/MessagesContainer.vue
@@ -19,7 +19,7 @@
     <div
       :class="[
         isMobileDrawerOpen ? '' : 'hidden',
-        'bg-yellow-100 w-screen sm:h-vh-2/3 md:h-vh-1/2 absolute sm:relative top-0 left-0 sm:flex-auto sm:block sm:border sm:border-gray',
+        'bg-white w-screen sm:h-vh-2/3 md:h-vh-1/2 absolute sm:relative top-0 left-0 sm:flex-auto sm:block sm:border sm:border-gray',
       ]"
     >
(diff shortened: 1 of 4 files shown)


src/components/organisms/MessagesContainer.vue
```diff
@@ -1,12 +1,13 @@
 <template>
-  <div class="flex sm:h-vh-1/2">
+  <div class="flex md:h-screen">
```
nibivi77, line 2: h-screen is too big in my opinion. Maybe 3/4 would be better?

src/components/organisms/MessagesContainer.vue
```diff
@@ -1,12 +1,13 @@
...
     <div
       class="
         w-full
         sm:w-2/5
         md:w-1/3
         sm:max-w-sm sm:border sm:border-r-0 sm:border-gray
         flex flex-col flex-shrink-0
+        px-1
```
nibivi77, line 10: The px-1 does not reflect the Figma design so much and it creates a small white margin between the inbox box and the inbox items.
image

shewood: I know,  but I wanted to see as on mobile it was getting clipped.

src/components/organisms/MessagesContainer.vue
shewood: maybe 3/4 would be better. what size do you find it too big?

nibivi77: On mobile at 320px width, on landscape it looks like this:

nibivi77 (approved):
This looks good. Tested on multiple screen sizes.

PR #40, opened 2021-06-22
added tabindex

TishamIslam:
## Description 
Added focus to the message list and conversation window, so you can scroll through them using only keyboard controls.


nibivi77 (approved):
Good ! :)

PR #42, opened 2021-06-22
changed font colours and h1 to h3

TishamIslam:
## Description

Changed h1 tag in the message card to an h3 tag. I also changed font colours to be #252525 / rgb(37,37,37) for all the associated text.


shewood (approved):
Looks good.

PR #43, opened 2021-06-22
chat messages returned in one call by id

DaceyTom2:
Description

Changed chat messages to come from one call in order by ID

### DIFF
--- a/src/components/organisms/ConversationWindow.vue
+++ b/src/components/organisms/ConversationWindow.vue
@@ -53,7 +53,7 @@ export default {
 
     function sendMessage(msg) {
       const newMessage = {
-        chatId: chatMessage.value.chatId,
+        id: chatMessage.value.id,
         isUser: true,
         text: msg,
       };
--- a/src/components/organisms/ConversationWindow.vue
+++ b/src/components/organisms/ConversationWindow.vue
@@ -2,20 +2,20 @@
   <div class="w-full h-full flex flex-col p-4 sm:p-0">
     <message-header
       backIcon="Back"
-      imageName="VC"
-      altText="Virtual Concierge"
-      headerText="Virtual Concierge"
+      :imageName="chatMessage.senderIcon"
+      :altText="chatMessage.senderIconAltText"
+      :headerText="chatMessage.senderName"
     />
     <div class="flex h-full overflow-auto">
       <img
-        src="../../assets/UserAccount/VirtualConcierge.svg"
-        alt="Bot Image."
+        :src="icons[chatMessage.senderIcon]"
+        :alt="chatMessage.senderIconAltText"
         class="h-6 mt-auto w-10 mb-3"
       />
       <div class="w-full pr-6 overflow-auto flex flex-col-reverse">
         <ConversationMessage
-          v-for="message in chatbotConvo"
-          :key="message.chatMessageId"
+          v-for="message in chatMessage.messages"
+          :key="message.id"
           :isUser="message.isUser"
           :text="message.text"
         />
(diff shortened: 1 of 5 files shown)


src/components/organisms/ConversationWindow.vue
```diff
@@ -43,23 +43,26 @@ export default {
...
+    const chatMessage = computed(() =>
+      store.getters["chatMessages/getChatMessageByIdOrderedByMessagesDate"](
+        selectedInboxItemId
+      )
     );
+
     function sendMessage(msg) {
       const newMessage = {
-        chatId: chatbotObject.value.chatId,
+        chatId: chatMessage.value.chatId,
```
nibivi77, line 56: This will need to be changed to id instead of chatId

src/store/modules/chatMessages.js
```diff
@@ -21,13 +21,12 @@ const getters = {
...
     }
-    return null;
+    return cm;
   },
-  getChatMessageObject(state) {
-    // state[0] will change to the selected inbox item
-    return state.chatMessages[0];
+  getChatMessageById: (state, getters) => (id) => {
+    let cm = state.chatMessages.find((chatMessage) => chatMessage.id === id);
+    return cm;
```
nibivi77, line 29: this function could be changed to:
`getChatMessageById: (state) => (id) => state.chatMessages.find((chatMessage) => chatMessage.id === id) || {}`

This way, we return an empty object in the case of undefined. (its also short making it look like we know what we are doing)

src/store/modules/chatMessages.js
```diff
@@ -7,10 +7,10 @@ const state = {
 // getters
 const getters = {
   // Get virtual concierge items
-  getChatMessagesOrderedByDate(state) {
-    if (state.chatMessages.length !== 0) {
-      // [...state.chatMessages[0].messages] makes a copy of the state array. Getters cannot mutate the state directly.
-      const conversation = [...state.chatMessages[0].messages].sort((a, b) => {
+  getChatMessageByIdOrderedByMessagesDate: (state, getters) => (id) => {
+    let cm = { ...getters.getChatMessageById(id) };
+    if (cm.messages.length !== 0) {
```
nibivi77, line 12: If cm is not what we expect, cm.messages will crash.

We need to check before the `if (cm.messages.......)` for the content of cm. 

Maybe add a `if (!cm) return {}`

or if cm returns as an empty object if nothing is found - (see comment below) - we could add this check:
`Object.keys(cm).length === 0`

nibivi77 (changes requested):
This is starting to shape like a good responsive application. This PR will show its true potential with the help of your next PR.

src/store/modules/chatMessages.js
DaceyTom2: I put in a check to see if cm is undefined. If it isn't it will escape with the existing object which is still undefined

src/store/modules/chatMessages.js
DaceyTom2: I don't want to return an empty object because it is harder to handle then checking for undefined. I did clean up the syntax

src/components/organisms/ConversationWindow.vue
DaceyTom2: Changed (In multiple places)

src/store/modules/chatMessages.js
nibivi77: This is tottally personal preferences but you dont need to have the `return` key word or the `{` `}` brackets after the fat arrow.

nibivi77 (approved):
Good work! I left 1 comment about the syntax. Just a cool syntax trick (I find personally)

src/store/modules/chatMessages.js
DaceyTom2: Personally I find it more readable with brackets and the return but am fine with it either way

PR #45, opened 2021-06-22
Emails can be returned by id from inbox section

DaceyTom2:
Description

Inbox items are derived from chatMessages and emails
emails can be selected based off of id selected
emails return full object with list of messages

TODO: 
teaser text and dayRead need to be generated dynamically
When a message is viewed the read indicator needs to be updated


src/store/modules/emails.js
```diff
@@ -11,18 +10,16 @@ const getters = {
...
-      (sender) => sender.senderId === state.selectedSenderId
-    );
+    return getters.getMailObject.find((sender) => sender.id === id);
   },
   //returns all the emails, with most recent ones first
-  getEmailsOrderByDateAsc(state, getters) {
+  getMailObjectEmailsOrderByDateAsc: (state, getters) => (id) => {
     //get list of emails
-    const mail = getters.getSelectedSenderObject.emails;
+    var mailObject = { ...getters.getMailObjectById(id) };
```
nibivi77, line 20: Ewww a `var`! Get it out ! If only a `let` was here to save us!

src/store/modules/emails.js
```diff
@@ -11,18 +10,16 @@ const getters = {
...
     //sorts a copy of the list of mail (getters should not mutate state)
-    const sortedMail = [...mail].sort((a, b) => {
+    mailObject.emails = [...mailObject.emails].sort((a, b) => {
```
nibivi77: Maybe we should have some error handling here just like the chatMessages.js?

What if mailObject is undefined?

src/store/modules/inbox.js
```diff
@@ -2,34 +2,38 @@
...
+      });
+    });
+
+    rootGetters["emails/getMailObject"].forEach((item) => {
+      inboxItems.push({
+        id: item.id,
+        senderIcon: item.senderIcon,
+        senderName: item.senderName,
+        teaserText: "",
+        dayRead: "wed",
```
nibivi77, line 32: If we can, we could call a function, pass the item, and return the date we need. Just an idea. (This can be implemented in a different PR)

P.S.: The same applies for teaserText.

nibivi77 (changes requested):
Looking good Tom!

src/store/modules/emails.js
DaceyTom2: It's gone... Pleassseee forgive me,I learned js so long ago

src/store/modules/inbox.js
DaceyTom2: Yes, I didn't want to jump into anything complex before establishing a baseline. Once we get this PR merged that can be done

src/store/modules/emails.js
DaceyTom2: I must have added it in while you were commenting on this PR. However! I did add a check to make sure some emails existed like in chatMessages

src/store/modules/inbox.js
nibivi77: ok sounds good to me.

nibivi77 (approved):
This looks good to me!

PR #46, opened 2021-06-22
Moved SkipNav into header + added label to SkipNav

TishamIslam:
## Description
This moved the SkipNav, which was outside all [page landmarks](https://dequeuniversity.com/rules/axe/4.2/region?application=AxeChrome) (headers, footers, and other sections), into the very top of the header so that it complied with accessibility guidelines/regulations. 

I also added a aria-label for the skip nav, I'm not 100% sure if that's appropriate/necessary for the component though.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -1,5 +1,8 @@
 <template>
   <div id="app">
+    <!-- skip nav component -->
+    <the-skip-nav />
+
     <!-- The header component -->
     <the-header />
 
@@ -22,6 +25,7 @@
 <script lang="ts">
 import TheHeader from "./components/molecules/TheHeader.vue";
 import TheFooter from "./components/molecules/TheFooter.vue";
+import TheSkipNav from "./components/molecules/TheSkipNav.vue";
 import FeedbackButton from "./components/atoms/FeedbackButton.vue";
 
 export default {
@@ -30,6 +34,7 @@ export default {
     TheHeader,
     TheFooter,
     FeedbackButton,
+    TheSkipNav,
   },
 };
 </script>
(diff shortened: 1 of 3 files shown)


DaceyTom2 (approved):
Looks good

nibivi77 (commented):
I do not know if we should have the skip-nav in the header.
On https://www.canada.ca/en.html, its above the header.

The goal of this component is to allow screen readers to skip to the desired sections of the application. 

Since we do not have enough information of this website, I honestly think we don't need this component. What do you think @TishamIslam @DaceyTom2 @shewood ?

DaceyTom2:
I was under the impression it was for navigating around canada.ca. I could be wrong but I wouldn't remove it unless requested by Thomas. I know there were some conversations that went on about it being included previously.

TishamIslam:
Now that Ben mentioned canada.ca, I see that they contained the SkipNav in a nav section so I think I'll just copy that instead.

nibivi77:
> I was under the impression it was for navigating around canada.ca. I could be wrong but I wouldn't remove it unless requested by Thomas. I know there were some conversations that went on about it being included previously.

Fair enough, we can keep it, its not big. The only thing we should do (not in this PR), is to make it functional. Because right now its leading to a div that does not exist with the id `#maincontent`.

But yes, it is designed to be at the top of the page, invisible, so that its the first control a screen reader "lands" on. From there, the user/screen reader can decide to skip to specific sections of the page. (Instead of hitting tab, tab, tab, tab... until they get to where they want to go).

nibivi77:
> Now that Ben mentioned canada.ca, I see that they contained the SkipNav in a nav section so I think I'll just copy that instead.

Good idea!

PR #48, opened 2021-06-23
Vc 151 added top part to main

TishamIslam:
## Description 
Added some vertical padding to the inbox and message headers so that they don't seem too crammed.

### DIFF
--- a/src/components/molecules/Inbox.vue
+++ b/src/components/molecules/Inbox.vue
@@ -1,5 +1,5 @@
 <template>
-  <h1 class="font-heading font-bold text-4xl pl-3 pt-4">Inbox</h1>
+  <h1 class="font-heading font-bold text-4xl pl-3 pt-4 pb-4">Inbox</h1>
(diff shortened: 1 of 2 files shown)


src/components/molecules/Inbox.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <h1 class="font-heading font-bold text-4xl pl-3 pt-4">Inbox</h1>
+  <h1 class="font-heading font-bold text-4xl pl-3 py-4">Inbox</h1>
```
DaceyTom2, line 2: In the design the word Inbox sits lower in the div. Almost resting on the bottom of it

image

DaceyTom2 (changes requested):

src/components/molecules/Inbox.vue
TishamIslam: Added more padding to the top and less to the bottom. Let me know if there's anything else that you've noticed

DaceyTom2 (approved):

PR #49, opened 2021-06-23
updated the alert icon

shewood:
[vcon-150](https://decdvirtualconcierge.atlassian.net/jira/software/projects/VCON/boards/2?selectedIssue=VCON-150)
Placed the alert icon in the center as per figma.


nibivi77 (approved):
I prefer this than what it was before :) Good !

PR #50, opened 2021-06-23
Vcon 135 rectangular messages

TishamIslam:
## Description
Just added a minimum width in tailwind and applied it the chat messages + centered the text so it didn't look weird. The rem value looks a little strange, but as far as I can tell, the standard minimum size is equivalent to a 3-letter message or so on most chat applications (tested on Instagram and on iMessages) which I implemented here as well.

## Note 
Tailwind doesn't have any default minimum sizes aside from 0, full, and content-based min/max, so I added a custom minimum size for the messages. This could be easily added for other parts of the application.

### DIFF
--- a/tailwind.config.js
+++ b/tailwind.config.js
@@ -57,6 +57,9 @@ module.exports = {
         "vh-1/3": "calc(100vh / 3)",
         "vh-1/4": "calc(100vh / 4)",
         "vh-1/5": "calc(100vh / 5)",
+      },
+      minWidth: {
+        "msg": "3.5rem"
       }
     },
   },
--- a/tailwind.config.js
+++ b/tailwind.config.js
@@ -59,7 +59,7 @@ module.exports = {
         "vh-1/5": "calc(100vh / 5)",
       },
       minWidth: {
-        "msg": "3.5rem"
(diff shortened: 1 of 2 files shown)


tailwind.config.js
```diff
@@ -57,6 +57,9 @@ module.exports = {
         "vh-1/3": "calc(100vh / 3)",
         "vh-1/4": "calc(100vh / 4)",
         "vh-1/5": "calc(100vh / 5)",
+      },
+      minWidth: {
```
shewood, line 61: I don't think you need to add anything special. Tailwind has min-width utils already

tailwind.config.js
```diff
@@ -57,6 +57,9 @@ module.exports = {
...
+        "msg": "3.5rem"
```
shewood, line 62: can you make it so that it is less specific for the message, so something like "3.5": "3.5rem" ?

TishamIslam: Wouldn't a "3.5" class not work because of the period? If not, then I suppose a "7/2" class would also work (even if it looks a little weird haha)

shewood (approved):
I approve.

PR #51, opened 2021-06-23
Vcon 74 message box component

DaceyTom2:
Description

All of inbox is derived from chatMessages and email.
1st chatMessage is selected by default
Messages are read on viewing


shewood (commented):
this looks very cool to see the text show up in the inbox. Is the name of the message sender supposed to update too? or is that another task?

DaceyTom2:
@shewood I believe the sender is always supposed to be the person or bot you are having a conversation with. I'll check with Thomas at standup

shewood (approved):

PR #52, opened 2021-06-23
added margins to conversation messages

TishamIslam:
## Description 
Fixes the issue of there being too much overlap between messages on small screen sizes. Image below shows the amount of margin that the messages will have on each side (40px).

image


shewood (approved):
It looks like this should work! Nice work.

PR #56, opened 2021-06-25
Vcon 60 loading state

DaceyTom2:
Description

For this PR I have the loading state forced. It will have to be removed after Thomas reviews the design and before merging.


neilmispelaar:
This works perfectly for the demo - so not recommending any changes - my only feedback is that the check on `v-if="inboxItems.length === 0"` to determine loading state would be risky in a more robust environment because what if an inbox really did only have 0 items (ah the dream of Inbox Zero - I have yet to experience thee).

A more robust approach // albeit require more code, would be to have the vuex store keep a `loading` state flag and then when the app is initialized the loading flag is set to true, and then when all the vuex store init functions are completed the loading state is set to false. 

Components then could do something like:
 
```
<template>
<div v-if="loaded === false">
...
</div>
<div v-else>
...
</div>
</template>
<script>
import { reactive } from 'vue';
...
setup() { 

  // Setup local reactive elements
  const localState = reactive({
    loaded: null,
  });
  
  localState.loaded = computed(() => store.getters["inbox/getLoadingState"]);

  return { 
  ...toRefs(localState),
  }
}
</script>
```

This could further be enhanced with an `error` state as well, but that's super overkill for now.

neilmispelaar:
My only other feedback is that - it would be cool to be able to use tailwindcss `annimate-pulse` class but I'll let Thomas Hoy pipe in on design

https://user-images.githubusercontent.com/9325038/123473485-a09fcc00-d5c6-11eb-96d1-a07825ef5bf5.mov

TishamIslam (commented):
It's looking really good so far! I don't think there's any issues from what I can tell, but I'd like to give it a more thorough review and do a bit of local testing before approving this.

neilmispelaar:
This looks amazing! 

Love it all. I know I'm not on the approval list but it looks great to me!

One opportunity to improvement for code readability in `Inbox.vue`:

- Separate the loading presentation markup from the actual data populated area 
- Someone quickly reading the code might miss that the `h1` is written just after a loading state check
- Organizing the code into two blocks (one for loading, and one for loaded) would help when someone is scanning the code
- In the future adding an error state might also be easier because the different views are separated 

```
<template>
  <!-- Loading State --> 
  <div v-if="!inboxLoaded">
    ... all loading markup ...
  </div>

  <!-- Loaded State --> 
  <div v-else>
    ... all loaded markup ...
  </div>
</template>
```

In the future we can also look at adding the aria roles I'll make a separate ticket for that because: complicated

TishamIslam:
Aside from what Neil mentioned above, I think it's perfect!

TishamIslam (approved):

shewood (changes requested):
I love the animation. it looks awesome.
I am concerned that there is no content though and once this gets merged the app will no longer function? or will you add a fix just before merging?

DaceyTom2:
@shewood I have initialise temporarily commented out but will remove before merging. This is so we can view changes on netlify

DaceyTom2:
@neilmispelaar Index is updated to be more readable. Good thinking!

shewood (approved):
Looks great!

PR #57, opened 2021-06-25
try to fix border thickness

TishamIslam:
## Description
Quick fix to solve the input field issue, not 100% sure if it works on all mobile platforms but the deploy preview on my iPhone 7 
no longer has the weird extra-thick line on it.
[Source](https://stackoverflow.com/questions/54763835/how-to-avoid-an-extra-border-top-thickness-on-bootstrap-input-field-in-iphone)


DaceyTom2 (changes requested):
This can be done using tailwind.

Checkout https://tailwindcss.com/docs/background-clip

TishamIslam:
Lol thought it was too specific of a thing to be included in tailwind... guess I shouldn't make any assumptions :P

DaceyTom2:
Haha I would have thought as well. If it wasn't for your pr and my confusion on the subject I would have assumed no

DaceyTom2 (approved):
Beautiful

PR #58, opened 2021-06-25
added basic pr template to root

TishamIslam:
###Description 

Just a simple PR template to speed the process up a little, if we can add more to it, then that's great!


DaceyTom2 (approved):

neilmispelaar:
Just curious - is there a standard practice on where to put these files - or are they normally dropped into the `root` folder?

PR #59, opened 2021-06-25
Vite remote host as npm script

TishamIslam:
### Description
Adds a little npm script so that you can run your dev server on devices aside from your computer, which is very useful for testing mobile versions of the site alongside the desktop version. All you need to do is run `npm run dev:host`


DaceyTom2 (approved):
This is a good nice to have!

PR #62, opened 2021-06-28
aligned feedback button to inbox on all sizes

TishamIslam:
### Description
I aligned the feedback button to the site banner in the mobile (xxs-xs) view.
I also re-sized and aligned the feedback button to the inbox for the xs-sm view, where previously it was as wide as the entire main content area. 

### Additional Notes
I feel like there's a better way to do this, as what I did involved moving the feedback button into its own separate container so that it could have the right margins on the mobile view.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -7,7 +7,7 @@
     <the-header />
 
     <!-- Page content -->
-    <div class="w-full mx-auto container pb-3 sm:p-3">
+    <div class="w-full pb-3 sm:p-3">
       <main>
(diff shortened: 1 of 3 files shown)


shewood (commented):
This seems to work, The other option you can look at is to change the size in the FeedbackButton.vue file to this class. 
w-full rounded justify-center py-2 mt-6 sm:w-2/5 md:w-1/3 md:max-w-sm

shewood (approved):
This looks good.

PR #64, opened 2021-06-28
increased margins for the inbox on mobile

shewood:

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -7,7 +7,7 @@
     <the-header />
 
     <!-- Page content -->
-    <div class="w-full mx-auto container pb-3 sm:p-3">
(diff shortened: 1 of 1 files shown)


src/App.vue
```diff
@@ -7,7 +7,7 @@
     <the-header />
 
     <!-- Page content -->
-    <div class="w-full mx-auto container pb-3 sm:p-3">
+    <div class="w-full px-6 sm:mx-auto sm:container pb-3 sm:p-3">
```
nibivi77, line 10: In the Figma mockups, the px is set to 16px:
image

To get this 16px we can use px-4 insteaad of px-6.

nibivi77 (changes requested):
Good catch but I think we are slightly off. See comments below for the x-axis adding.

shewood:
Closing this pr in favor of another one with the same change.

PR #66, opened 2021-06-28
adjusted margins and paddings in conversation window

TishamIslam:
### Description
Modified the margins such that the VC icon would have an equal amount of space on either side of it.

### Additional Notes
Also added some vertical padding as I had noticed that there was some in the figma but not on the site.

### DIFF
--- a/src/components/organisms/ConversationWindow.vue
+++ b/src/components/organisms/ConversationWindow.vue
@@ -20,6 +20,7 @@
           flex flex-col-reverse
           font-body
           text-gray-dark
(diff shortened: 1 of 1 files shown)


shewood (changes requested):
I just made the change for py-px and px-2 as requested by Thomas as part of another task. 
this is for the little icon, it's in another part no?

TishamIslam:
Closing this PR as Sheila implemented the changes listed here within her PR just before/during a Teams call with Thomas and I. (#68)

PR #67, opened 2021-06-29
Made mobile VC conversation header + text input sticky 

TishamIslam:
### Description
The header and input box are now sticky on the mobile view. 
Also added a bit of transparency (95%) to the header. 

### Additional Notes
Note that this did _not_ make the header on the email view sticky nor transparent/make that specific header component sticky everywhere else. Will change if that is preferred.

### Issues
I can't seem to get it to unsticky when it goes into desktop view. There is no visible issue except on landscape mode for certain mobile devices where the faulty sticky behaviour can be seen. Simply adding a breakpoint class like `sm:relative` to override the sticky class doesn't work. Not sure how else to get around it


nibivi77 (changes requested):
This is looking really good but watch out for teh styling of the sticky elements.
image

The bottom text input and buttons are not taking the full width.

shewood (approved):
This looks good. 
If you change the w-auto to w-full on the TextInput it fixes the short bar. 
The bg-clip-padding shows the background as the padding but in this case it allows the bubble to show through as well.  
image

nibivi77 (approved):
Nice flow I like it :+1:

PR #69, opened 2021-06-29
Vcon-164 removed the default selected state

shewood:
### Description
removed the default focused state
added active state for the inbox items.

### DIFF
--- a/tailwind.config.js
+++ b/tailwind.config.js
@@ -64,6 +64,11 @@ module.exports = {
       }
     },
   },
-  variants: {},
+  variants: {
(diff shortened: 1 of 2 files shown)


tailwind.config.js
```diff
@@ -64,6 +64,11 @@ module.exports = {
       }
     },
   },
-  variants: {},
+  variants: {
+    extend: {
+      backgroundColor: ['active'],
+      borderRadius: ['active'],
```
TishamIslam, line 70: I don't see this bit being used.

PR #70, opened 2021-06-30
Vcon 180 connection error handler

nibivi77:
### Description
Converted the calls to the service as Promises.
If there is an error connecting, The error message will be displayed from the bot.

### Additional Notes
 - After the connection has been made with the bot, if an error occurs in the connection, it will be console logged.
 - If there is a place where we can improve in the code, let me know! (ex.: error checks, code logic, etc...)


src/store/modules/chatMessages.js
```diff
@@ -36,47 +36,61 @@ const getters = {
 // actions
 const actions = {
   initializeChatMessages({ dispatch, commit, rootGetters }) {
-    const conversationId = "VC01";
-    let directLineMessageRecievedHandler = (userName, message) => {
+    let directLineMessageRecievedHandler = (userName, message, convoId) => {
+      if (!state.chatConversation.some((chat) => chat.id === convoId)) {
+        commit("addChatConversation", {
```
DaceyTom2, line 41: It's not a good idea to create chat conversations on message sent. It sends us further down the rabbit hole of being locked into having an init send message

src/store/modules/chatMessages.js
```diff
@@ -36,47 +36,61 @@ const getters = {
...
+        ],
+      });
+    });
   },
 
   async sendChatMessage({ commit, rootGetters }, payload) {
     // make await api call here with the text
     //TODO: get username
-    ChatMessageService.sendMessage(payload.message, "userName");
+    ChatMessageService.sendMessage(payload.message, "userName")
```
DaceyTom2, line 87: This looks really solid and makes it easier to implement the service

src/store/modules/chatMessages.js
```diff
@@ -127,6 +141,13 @@ const mutations = {
       messages: payload.messages,
     });
   },
+  // setChatConversationId(state, payload){
```
DaceyTom2, line 144: Does this code work and is it needed? Can probably be removed

src/store/modules/chatMessages.js
```diff
@@ -36,47 +36,61 @@ const getters = {
...
+    });
   },
 
   async sendChatMessage({ commit, rootGetters }, payload) {
     // make await api call here with the text
     //TODO: get username
-    ChatMessageService.sendMessage(payload.message, "userName");
+    ChatMessageService.sendMessage(payload.message, "userName")
+      .then((res) => {
+        console.log("sending message response: ", res);
```
DaceyTom2, line 89: IS it possible to return username, message and convoId here? This could clean up the code by not having to pass the directLineMessageRecievedHandler function in. We can just call directLineMessageRecievedHandler(username, message, convoId) here to add messages

DaceyTom2 (changes requested):

src/store/modules/chatMessages.js
DaceyTom2: It is unlikely this can be done

src/store/modules/chatMessages.js
DaceyTom2: Can get conversation id by subscribing to connectionStatus and responding to a status of 2 meaning connected

DaceyTom2 (approved):

PR #71, opened 2021-06-30
VCON 148 Send icon states

TishamIslam:
### Description
Added states to the send icon:
- when there's any text inputted, the icon will update to be active
- if there is no text (text deleted/sent) the icon will be in its inactive state. ~~it can still be focused upon, but not used~~ can no longer be tabbed to if there is no text
- upon tabbing to/hovering over the button, if there's text in the input, the send icon will transition to the focused state
- tabbing away/hovering out will return the button to the active state
- hitting enter while focused/clicking the button will immediately send the message and set the icon to the inactive state

### Additional Notes
~~Focus state is not activated on click as of yet. Not sure if that's actually needed, will talk to Thomas about it.~~
Focus state activates on hover/tab.
~~Also the outline is still there, dunno how to fix that. Will probably ask Sheila or someone else about that.~~
Sheila has said that it is a non-issue as the focus border always appears on tabbing.

### DIFF
--- a/src/components/atoms/TextInput.vue
+++ b/src/components/atoms/TextInput.vue
@@ -6,7 +6,7 @@
     v-model="text"
     aria-label="Send message"
     @keyup.enter="sendText"
-    @change="checkSendBtnActive"
+    @input="checkSendBtnActive"
   />
   <div class="w-auto bg-gray-infolt text-gray-dark">
     <chat-option-button
--- a/src/components/atoms/TextInput.vue
+++ b/src/components/atoms/TextInput.vue
@@ -6,6 +6,7 @@
     v-model="text"
     aria-label="Send message"
     @keyup.enter="sendText"
+    @onchange="checkSendBtnActive"
   />
   <div class="w-auto bg-gray-infolt text-gray-dark">
     <chat-option-button
@@ -18,9 +19,12 @@
       class="float-right cursor-pointer"
       aria-label="Send message"
       @click="sendText"
+      @focus="onSendBtnFocus"
+      @blur="checkSendBtnActive"
     >
       <img
-        src="../../assets/Send/Active/Default.svg"
+        class="focus:border-none"
+        :src="icons[iconState]"
         alt="Send Image."
         aria-hidden="true"
       />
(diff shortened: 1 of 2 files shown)


src/components/atoms/TextInput.vue
```diff
@@ -41,21 +55,45 @@ export default {
   emits: ["add-message"],
   setup(_, context) {
     const text = ref("");
+    const iconState = ref("SendInactive");
+
+    //these refs are reassigned to their respective components on mounting
+    //as with other refs, the actual elements are accessed with .value (input.value, for example)
+    const input = ref(null);
+    const sendMsgBtn = ref(null);
```
DaceyTom2, line 63: Instead of creating a ref to the button we manipulate for tab index we can have a variable that we set. IE var sendMessageTabIndex. Then :index="sendMessageTabIndex"

src/components/atoms/TextInput.vue
```diff
@@ -15,21 +17,33 @@
...
     </button>
   </div>
 </template>
 <script>
+import icons from "../../assets/icons.js";
+
 import ChatOptionButton from "./ChatOptionButton.vue";
 import { ref } from "vue";
+
+//let iconState = "SendInactive"
```
DaceyTom2, line 45: remove this

src/components/atoms/TextInput.vue
```diff
@@ -15,21 +17,33 @@
       @send-button="sendBtnText"
     />
     <button
+      ref="sendMsgBtn"
       class="float-right cursor-pointer"
       aria-label="Send message"
+      tabindex="0"
       @click="sendText"
+      @focus="onSendBtnFocus"
+      @mouseover="onSendBtnFocus"
```
DaceyTom2, line 26: When I hover over with my mouse I don't get the focus button

DaceyTom2 (changes requested):

src/components/atoms/TextInput.vue
TishamIslam: Did you write anything into the textbox? It won't display the focus icon unless there is text/the button is active

src/components/atoms/TextInput.vue
TishamIslam: Right, that makes more sense.

src/components/atoms/TextInput.vue
DaceyTom2: Yes, I can show you what it looks like from my perspective when you have a minute

DaceyTom2: Hahaha nvm... I was in mobile mode so no mouse...

DaceyTom2 (approved):
Good stuff!

PR #72, opened 2021-07-02
VCON-189 Scroll messages on load + on message send

TishamIslam:
### Description
Scrolls messages when the chat window is initially loaded and when new messages are sent. Since it's done in the "onMounted" function, it also scrolls when it's clicked from the inbox. 

### Additional Notes
~~I feel like the way that I did this was sort of "hacky" and maybe not the best way to do it. From what I can tell, the scrollIntoView() function is the proper function to use but I feel like there's a better way to use it than I have.~~

### Issues
Doesn't seem to work on mobile.


neilmispelaar:
When you hit enter the whole window drops down (on Firefox - not sure if other browsers are behaving differently)

https://user-images.githubusercontent.com/9325038/124367081-4b953300-dc22-11eb-91ba-1f938ebcda85.mov

neilmispelaar:
Here is some code I used in an older project that may be helpful...

https://github.com/Toad-Labs/User-testable-chatbot-prototype/blob/main/src/components/ChatWindowMessages.vue

My strategy was to use `document.querySelector` to get a reference to a dom element, and then use the native javascript attributes to set the `scrolltop` 

```

// Cache the chat message block
 var chatMessagesContainer =  document.querySelector('.my-chat-messages');

chatMessagesContainer.scrollTop = chatMessagesContainer.scrollHeight;

```

One other item that may come up - `vue` takes a few microseconds to update the virtual dom, so if you update `scrollTop` too quickly the dom may not have the new message yet. If this comes up you can do something like the following to wait 50 milliseconds before scrolling down. 

```
// Scroll to the bottom of the container
    setTimeout(function() {
      chatMessagesContainer.scrollTop = chatMessagesContainer.scrollHeight;
    }, 50);
```

`Vue` has a `nextTick` function which basically does the same thing which waits for the dom to be updated - this is faster than waiting a blanket 50 milliseconds, but I can't remember how to use it.

TishamIslam:
> When you hit enter the whole window drops down (on Firefox - not sure if other browsers are behaving differently)
> 
>  drop.mov

Seems to do that on all browsers. I changed it to use scrollTop + scrollHeight like you mentioned and it works like a charm. I just noticed that neither method works properly on mobile, so that's an issue to resolve.

shewood (commented):
I tried it on firefox and also chrome and it didn't seem to scroll on message send.

nibivi77 (commented):
Tried on Chrome and did not seem to work for me. Any thoughts?

DaceyTom2 (approved):
Looks super clean and I love the speed

PR #73, opened 2021-07-05
Added skip nav functionality

TishamIslam:
### Description (Design Copied from [canada.ca](https://www.canada.ca/en.html))
Added skip nav functionality where the link skips to the inbox header (main content) 
Link appears when focused upon, otherwise can only be perceived with a screen reader
Clicking link focuses onto the inbox header; after tabbing away from the header, you cannot focus on the header again


shewood (changes requested):
This looks great. One small thing to change is the href to where the content is going. 
 <a href="#inbox-header">Skip to main content</a>

shewood (approved):
Looks Good

PR #75, opened 2021-07-06
bye bye jerry

shewood:
### Description
replaced Jerry with Job Bank
Removed the first message so only the job bank message appears as per task.


TishamIslam (approved):
Looks great!

PR #76, opened 2021-07-06
Vcon 216

shewood:
### Description
Removed the bottom padding on the overall container and because the button now has its own container, added margin to the top and bottom of the button container so that it would be even.


TishamIslam (approved):
Looks great, no issues here

PR #77, opened 2021-07-06
Reversed structural order of chat window content so that it is in order without css

TishamIslam:
### Description
Reversed the order of the elements in the conversation window (chat messages + the timestamp), and then modified the store message getter so that the order of the messages returned from the store would also be reversed. This way, everything is in the correct order in terms of the DOM/html structure. 

In short, there should be absolutely no visual differences in this PR unless you remove all css, in which case the basic markup should be identical to the actual display order.


shewood (approved):
Looks good. but add 
justify-end
so the convo starts at the bottom.

PR #78, opened 2021-07-07
Vcon 145 with click events

TishamIslam:
### Description
Removed the focus indicator for the input box on click


DaceyTom2 (approved):

DaceyTom2:
Not fond of adding and removing classes by variable but I did some research and couldn't figure a better way

TishamIslam:
Yeah, I'd much rather have a css-only solution but it doesn't look like there is one

PR #79, opened 2021-07-12
Vcon 174 vue i18n plugin

DaceyTom2:
### Description
Added i18n to project globally.
/src/locales has an en.json and a fr.json for English and French translations respectfully
By default the app is English but can change once splash page is implemented

### Additional Notes
Routes will be added in upcoming PR to toggle language in application
An example can be found in src/components/molecules/Inbox.vue of how to use


shewood (approved):
Looks good. I did as you said and changed the local to fr and it switched the one item to french.

PR #80, opened 2021-07-12
Various accessibility improvements

TishamIslam:
### Description
Covers several things, the biggest one being the conversation window (Applies to: VCON-134 WCAG-2.1.1 Keyboard): 
- Pressing enter or shift enter on the conversation window lets you access the conversation message by message
     - shift+enter lets you go directly to the latest message in your conversation
     - you can press these keys at any messsage to go to the beginning/end
- Pressing tab/shift+tab and up/down arrows lets you go back and forth between the messages
- Pressing escape or tab at the end lets you exit from the message navigation. 
    - There is hidden (until focused on) text at the bottom that tells you to do that. 
- Each of these have instructions attached to them via labels for screen readers. I am not 100% sure how to display that information to people that are using keyboard only but are not using a screen reader, however I don't think it's necessary since you can scroll with keyboard anyways. 
- The messages are untabbable/focusable unless you are navigating as described above, as they were previously. I've covered all edge cases that seemed reasonably likely (Like clicking away while focused on a message) but I might've added a bit of redundancy

Neil mentioned that the above changes might conflict with other wcag guidelines  

Misc. smaller changes:
Added some extra labeling for quick replies so you can be informed on what the text contains (VCON-124: WCAG 1.3.5 identify input purpose)  

Changed heading level from 3 to 2 + added screen-readable time stamps in messageCards. (VCON-120: WCAG 1.3.1 info and relationships, VCON-141: WCAG 2.4.6 headings and labels)

Added extra hidden labeling to the inbox so that it describes whether or not its contents are unread and selected/unselected. (I think I intended this to be part of VCON-119 (Non-text content) and 141 (labels) but it seems to apply more to [VCON-122](https://decdvirtualconcierge.atlassian.net/browse/VCON-122?atlOrigin=eyJpIjoiNjI4YjkxNDBjMzAzNGVhOTgyODY3NDI3YmZlYTllYzgiLCJwIjoiaiJ9) which @shewood is covering)

### Additional Notes
Apologies in advance, this ended up being bigger than it should've been. Feel free to call me so you can review it more quickly  and effectively. 
I am using the NVDA screen reader for Windows, I'd recommend using a screen reader that can also show what it is reading out for convenience. Please let me know if you tested anything with a different screen reader and found it inaccessible. 
This PR does _not_ cover these WCAG guidelines in full, that'd probably make it even bigger than it is right now.


shewood (changes requested):
When you first view the page, the inbox shows the chat as being unread and then right away, shows it as being read even though on mobile, I didn't click to read it yet.
Also, the whole page shifts to the left so half of the page is partially covered when you are navigating through the items using voice over.

DaceyTom2 (changes requested):

src/components/atoms/ConversationFooter.vue
```diff
@@ -0,0 +1,33 @@
...
+    class="sr-only focus:not-sr-only font-body"
+    :tabindex="tabindex"
+    @keyup.esc="returnToChatWindow"
+    @keydown.up.prevent="prevElement($event)"
+    @keydown.down.prevent="exitToInput"
+    @keydown.shift.tab.prevent="prevElement($event)"
+  >
+    You have reached the end of the conversation, press escape to exit to the
+    conversation window. Press down or tab if you want to send a message. Press
+    shift+tab to revisit the conversation, press enter to return to the top of
```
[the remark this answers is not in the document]
TishamIslam: Given that shift tabbing is default behaviour for focusing through elements backwards, I think it's fine as people that use screen readers would do that all the time.

src/components/atoms/ConversationTabbables.js
```diff
@@ -0,0 +1,12 @@
+export function ConversationTabbables(_, context) {
```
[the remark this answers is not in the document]
TishamIslam: Alright, copy-pasting code just doesn't sit well with me even when it's just a single line thing 😅. definitely was a lot more work to put it out in a separate file though, so I'll follow your advice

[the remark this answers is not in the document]
DaceyTom2: It will be easier to read. If the code was a bit more complex I'd be on board 100%

DaceyTom2 (approved):
Looks good!

src/components/molecules/Inbox.vue
```diff
@@ -12,17 +12,22 @@
   </span>
   <!-- Loaded State -->
   <span v-else>
+    <span :id="'inboxHeaderDesc'" class="hidden">
+      Inbox header, press tab or click to access inbox items
```
shewood, line 16: This doesn't work as expected when clicking enter here with NVDA

src/components/molecules/InboxItem.vue
```diff
@@ -15,35 +15,55 @@
         v-if="inboxItem.senderIcon"
         class="w-12 h-12 md:w-12 md:h-12"
         :src="icons[inboxItem.senderIcon]"
-        alt="Inbox icon"
+        alt="Inbox icon."
```
shewood, line 18: Is it possible to have an empty quote here for the alt text as this icon doesn't do anything, it is not clickable and inbox icon isn't relevant for the assistant and it isn't translated.

shewood (commented):
I am weary about the long strings if text that may or may not apply depending on the reader. 
Also, the text isn't set up yet for localization but they may be in another task that you have going on.

src/components/molecules/Inbox.vue
TishamIslam: This might've been a misunderstanding, it's supposed to mean that you tab to or click on the inbox items. If you can't access the label, try using the skip nav as the label can only be seen when you focus on it. 

As a side note, I'd also recommend looking into NVDA's Browse mode, it allows you to focus on most elements by going to the closest element of X type by pressing a letter e.g. pressing 'h' or shift+'h' lets you go to next/previous heading, and each number lets you go to a heading of that level e.g. pressing '1' makes you go to the closest h1

src/components/molecules/InboxItem.vue
TishamIslam: I'll leave it as is for now, if Neil says that we should/can then we can change it in a later PR. I agree with you, the sender name is enough and (in my opinion) if we do decide to mention the icons, then we should be giving a full description of the icon regardless, which isn't very practical in this case.

TishamIslam:
> I am weary about the long strings if text that may or may not apply depending on the reader.
> Also, the text isn't set up yet for localization but they may be in another task that you have going on.

I can't remember where/if we've already discussed this, but I think we were waiting on the accessibility folks to evaluate this before changing the alt text.

PR #81, opened 2021-07-12
Vcon 140 back button

shewood:
### Description
The click function didn't link correctly using the keyboard to take the user back to the inbox. 
Changed it to be a button.

### DIFF
--- a/src/components/atoms/MessageHeader.vue
+++ b/src/components/atoms/MessageHeader.vue
@@ -1,6 +1,6 @@
 <template>
   <div class="flex border-b border-gray-200 px-2 py-4">
-    <button @click="clickBack" aria-label="back to inbox" tabindex="0">
+    <button @click="clickBack" aria-label="back to inbox">
       <img
         :src="icons[backIcon]"
         :alt="altText"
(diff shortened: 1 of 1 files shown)


src/components/atoms/MessageHeader.vue
```diff
@@ -1,13 +1,13 @@
...
-      :alt="altText"
-      class="sm:hidden h-10 mt-auto w-10 mr-4"
-      @click="clickBack"
-      aria-label="close modal"
-      tabindex="0"
-    />
+    <button @click="clickBack" aria-label="back to inbox">
+      <img
+        :src="icons[backIcon]"
+        :alt="altText"
```
DaceyTom2, line 6: I know you didn't create this alt text but it would be nice to say something like "back icon"

DaceyTom2 (changes requested):

src/components/atoms/MessageHeader.vue
TishamIslam: Agreed, this is what it says on the preview right now when you click on the chat.
image

DaceyTom2 (approved):
Looks good!

PR #82, opened 2021-07-13
Vcon 220 i18n routes

DaceyTom2:
### Description
Added /en and /fr routes for English and french language
Language defaults to English
If language is not en or fr the router will throw 404 since it won't match on pattern

### Additional Notes
Home is the default page for paths /en & /fr
By default /en will only appear in url if specified in url
Only 3 translation keys have been added thus far
ChatBot and services will need lang passed for translations on their side

### DIFF
--- a/i18n.js
+++ b/i18n.js
@@ -1,7 +1,7 @@
 import { createI18n } from 'vue-i18n'
 import messages from '@intlify/vite-plugin-vue-i18n/messages'
 
-export default  createI18n({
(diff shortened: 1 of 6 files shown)


DaceyTom2:
@neilmispelaar Great idea, I will change the default route to be a splash page and create the splash page in the next pr.

shewood (approved):
This looks good. Both links change the language.

PR #84, opened 2021-07-13
vcon 217 mobile margins

shewood:
### Description
fixed the margins for mobile view so that the line goes all the way across.


DaceyTom2 (approved):
Looks sharp!

PR #86, opened 2021-07-13
button position = absolute top right

TishamIslam:
### Description
If there's lots of quick replies, the button used to get moved around past the edge, sometimes to the bottom right, etc.
This fixes that issue.

### Additional Notes
This is designed the way Thomas wanted it. 
I'll remove the additional reply options before merging, they're included right now for display purposes.


shewood (changes requested):
The positioning bit seems good. but I am not sure about the addition of all of the options.
image

shewood (commented):
And if you remove all of the extra options, the send button looks to be in the wrong position.

image

shewood (commented):
try putting relative in this div 
div class="flex flex-row bg-gray-infolt text-gray-dark relative"
and absolute in this one 
div class=" flex w-12 flex-shrink-0  absolute right-0 top-0"
And I think it will work as you are expecting.

TishamIslam:
> I am not sure about the addition of all of the options.
> image

Don't worry about it, I'll remove the extra options before merging. This was just to show that the intended functionality works. 

> And if you remove all of the extra options, the send button looks to be in the wrong position.
> 
> image

Good catch, on it.



> 
> 
> try putting relative in this div
> div class="flex flex-row bg-gray-infolt text-gray-dark relative"
> and absolute in this one
> div class=" flex w-12 flex-shrink-0 absolute right-0 top-0"
> And I think it will work as you are expecting.
> Although it might still go over the bubble.

Not sure what you mean here, the divs you are referring to have relative and absolute in them already.

DaceyTom2:
I'm not a fan of how it fills up the entire screen. Perhaps there could be a more, expand or scroll option?

shewood (commented):
I am not sure if it is just because you have the extra button showing but the yes button looses the focus (cursor pointer) after you click the extra button on mobile.

shewood (approved):
Maybe it has to do with the netlify preview icon on the screen but when you click yes 3 times then you can't click again but it only happens on mobile when the button is close to the netlify button.

TishamIslam:
Hmm, maybe we can ask Neil about moving the button around as it seems to be interfering a little bit.

neilmispelaar:
> Maybe it has to do with the netlify preview icon on the screen but when you click yes 3 times then you can't click again but it only happens on mobile when the button is close to the netlify button.

Quick test of localhost main branch shows that the error (click yes 3 times) doesn't seem to occur. So maybe it's the netlify deploy preview javascripts, maybe it's just bad luck, I'm honestly not sure. 

Here's the video evidence of it working fine (main on localhost)

https://user-images.githubusercontent.com/9325038/126710601-5106263d-3582-4cd1-929b-2c60233c1d6b.mov

PR #87, opened 2021-07-14
message header back button can focus

DaceyTom2:
### Description
When focus or mouse over of back button on mobile message header icon will change to focused state

### Additional Notes
You can tab to focus or mouse over


shewood (approved):
Works really well.

PR #88, opened 2021-07-14
Check put in place for loaded state

DaceyTom2:
### Description
A check has been added for loaded before initialising modules again


TishamIslam (approved):
Solved the bug like a charm!

PR #89, opened 2021-07-14
Added en to path on desktop

DaceyTom2:
### Description
Clicking English in TheHeader.vue will now take you to the English version of the app instead of the splash page


TishamIslam (approved):
Works great!

PR #90, opened 2021-07-14
send lang in locale to chatbot

DaceyTom2:
### Description
On sendMessage Directline will send lang of en/fr to the chatbot service

### Additional Notes
Unable to do proper testing as chatbot is yet to handle lang


shewood (approved):
I didn't try it but it looks logical.

PR #91, opened 2021-07-14
edits for vcon 143

shewood:
### Description
work for WCAG 2.5.3 https://www.w3.org/WAI/WCAG21/Techniques/general/G211.html
removed the tabindex on the language control as it is now an actual link and no longer needs it. 
changed the quick reply buttons to read the actual text on screen.
changed the text input to be the same as the actual text although it might not be needed but maybe hidden for some voice controls,. more to do here. 
Added a few translations.


src/components/atoms/ChatOptionButton.vue
```diff
@@ -12,7 +12,7 @@
       bg-white
       hover:bg-gray-infolt
     "
-    aria-label="Quick reply"
+    :aria-label="$t(text)"
```
DaceyTom2, line 15: Adding text in here without translation should be fine as the text will be ideally passed in translated as a prop. The yes/no options can be translated in ConversationWindow.vue line 44 ":buttonOptions="['Yes', 'No']""

src/components/atoms/TextInput.vue
```diff
@@ -34,7 +34,7 @@
       <img
         class="focus:border-none"
         :src="icons[iconState]"
-        alt="Send Image."
+        alt="Send icon"
```
DaceyTom2, line 37: This alt text can be translated

src/components/molecules/TheHeader.vue
```diff
@@ -20,8 +20,6 @@
           <router-link
             id="lang-toggle-full"
             class="px-5 py-1 font-body underline text-blue-link"
-            aria-label="Change language"
-            tabindex="0"
```
DaceyTom2, line 24: good catch

DaceyTom2 (changes requested):

src/components/atoms/ChatOptionButton.vue
shewood: this is so the voice recorder reads "yes' and 'no" otherwise yes and no are both read as 'quick reply" and that makes no sense.

src/components/atoms/TextInput.vue
shewood: it doesn't seem to be used but I can translate it anyway.

src/components/atoms/ChatOptionButton.vue
DaceyTom2: The text itself doesn't need to be translated though

shewood: Right, of course.

src/components/atoms/TextInput.vue
shewood: Done

src/components/organisms/ConversationWindow.vue
```diff
@@ -41,7 +41,7 @@
       <!-- The logic on how the buttonOptions are passed as props will
                    depend on how we get the possible answers from VC. -->
       <TextInput
-        :buttonOptions="['Yes', 'No']"
+        :buttonOptions="[$t('yes'), $t('no')]"
```
DaceyTom2, line 44: Looks good!

DaceyTom2 (approved):

PR #92, opened 2021-07-15
Added lang to elements missing attribute

DaceyTom2:
### Description
French elements on English screens and English elements on French screens now have lang="respectiveLang"

### DIFF
--- a/src/views/Splash.vue
+++ b/src/views/Splash.vue
@@ -64,6 +64,7 @@
             role="button"
             draggable="false"
             to="/fr"
+            lang="fr"
             >Français</router-link
           >
         </div>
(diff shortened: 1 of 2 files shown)


src/views/Splash.vue
```diff
@@ -64,6 +64,7 @@
             role="button"
             draggable="false"
             to="/fr"
+            lang="fr"
```
TishamIslam, line 67: Github doesn't let me comment on the specific area (line 40 of this file) but you're missing a lang="en" for the English button.

TishamIslam (changes requested):

src/views/Splash.vue
DaceyTom2: Added lang=en to line ~40 and 103. Good catch!

TishamIslam (approved):

PR #94, opened 2021-07-16
left aligned chat option/ removed center align chat messages

TishamIslam:
### Description
Left-aligned all the chat-related items in the conversation window, as Thomas requested.

### Additional Notes
I also added break-words to the chat options, in the unlikely event that something gets too long for it to be displayed in one line.


DaceyTom2 (approved):
Looks good

PR #95, opened 2021-07-19
icon conditionally shows in conversationMessage

TishamIslam:
### Description
Moved the icon to each conversationMessage and conditionally rendered it based on whether it was the last element or not.
The alignment was to Thomas' specifications.


DaceyTom2 (approved):
Very clever!

PR #97, opened 2021-07-20
consistent cross-browser focus + placeholder colour

TishamIslam:
### Description
Added an explicit base focus styling to all elements instead of relying on browser defaults (resolves Firefox focus styles being hard to see/invisible). 
Explicitly defined the placeholder text colour in the textinput for sending messages to the VC (soon to be VA). This placeholder colour is the same as in the search box for consistency.


DaceyTom2 (approved):
Looks way better than my FF defaults!

PR #98, opened 2021-07-20
removed chatMessages.json

TishamIslam:
### Description
Removed chatMessages.json since it is no longer used.


DaceyTom2 (approved):
Reasonable to get rid of at this point

PR #99, opened 2021-07-20
redundant div with app id has been terminated 

TishamIslam:
### Description
I noticed that the div in App.vue wasn't actually needed since it seems like Vue makes the "#app" div on its own. This resolves a violation that was automatically detected with the Axe devtools extension.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -1,7 +1,5 @@
 <template>
-  <div id="app">
-    <router-view />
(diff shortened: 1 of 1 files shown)


DaceyTom2 (approved):
Interesting, I never noticed this being repeated... I'm surprised we never ran into issues with 2 elements with the same id

PR #100, opened 2021-07-20
added case for no messages in conversation

TishamIslam:
### Description
The blue notification icon that indicates “your message is unread” now disappears on page load automatically since the conversation is auto-selected. This was caused by not checking if a conversation was empty before assigning it a read date, so a separate check has been added for the case in which an empty conversation has a read time but no messages.

### DIFF
--- a/src/store/modules/inbox.js
+++ b/src/store/modules/inbox.js
@@ -16,18 +16,23 @@ const getters = {
       const lastMessage = messagesList[messagesList.length - 1];
 
       //dayRead will be unread if no value provided. The design currently accounts for only the weekday to be displayed
-      let dayRead = undefined;
+      let dayRead;
+      const lastReadDate = item.lastRead ? new Date(item.lastRead) : null;
+      const lastMessageReceivedDate = lastMessage?.receivedTime
+        ? new Date(lastMessage.receivedTime)
+        : null;
+
       if (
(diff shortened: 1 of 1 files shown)


nibivi77 (commented):
This does the job as expected! Good work! check out my comment :)

src/store/modules/inbox.js
```diff
@@ -28,6 +25,11 @@ const getters = {
         dayRead = new Date(lastMessage.receivedTime).toLocaleDateString("En", {
           weekday: "short",
         });
+      } else if (item.lastRead !== null) {
+        //to account for empty conversation/inbox item, so it doesn't remain always unread
+        dayRead = new Date(item.lastRead).toLocaleDateString("En", {
```
nibivi77, line 30: Instead of instantiating the Dates twice (one in the if statement and the other in the if/else clauses), would there be a way to declare it once?

Lines 23, 25 and 30

nibivi77 (commented):
This does the job as expected! Good work! check out my comment :)

src/store/modules/inbox.js
TishamIslam: Good catch, I'll be right on it!

nibivi77 (approved):
This is nice Good job!

PR #101, opened 2021-07-21
no tabbing to back button on desktop view/hidden

TishamIslam:
### Description
The back button that shows up when you open the drawer for messages/a conversation is no longer tabbable when you're on desktop (and is therefore hidden). Pretty sure this relates to one ore more of the WCAG guidelines, I'll update this and look for the relevant guideline(s) if there are any.


nibivi77 (approved):
That does the trick! Good catch!

PR #102, opened 2021-07-21
use focus-visible + remove input focus states

TishamIslam:
### Description
Removed focus states for inputs at Thomas Hoy's suggestion, used focus-visible tailwind styles so that focus states would activate only on tab/keyboard usage

### Additional Notes
~~Polyfills for other browsers incoming!~~
I think I implemented polyfilling for non-moz/chrome/edge; test with safari if avaiable. If it doesn't work, the plugins ([focus-visible](https://github.com/WICG/focus-visible) and [postcss-focus-visible](https://github.com/csstools/postcss-focus-visible)) have been included and I'm not 100% sure how to do it with Tailwind as [the docs ](https://tailwindcss.com/docs/hover-focus-and-other-states#focus-visible) aren't too descriptive, so let me know and we can take a crack at it together.

### Hoy Seal of Approval 
image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,6 +23,8 @@
         "@vitejs/plugin-vue": "^1.2.3",
         "@vue/compiler-sfc": "^3.1.2",
         "autoprefixer": "^10.2.6",
+        "focus-visible": "^5.2.0",
+        "postcss-focus-visible": "^5.0.0",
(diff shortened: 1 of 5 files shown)


DaceyTom2 (approved):
I tried it out on chrome and ff and it works really well. I'm unsure how to do something like this with tailwind

PR #103, opened 2021-07-22
Inbox uses ul and li instead of divs

DaceyTom2:
### Description
Instead of using divs within divs for the list of inbox items. Semantically we can use an ul(unorganised list) and create inbox items as li(list items). This is to help with accessibility

### DIFF
--- a/src/assets/styles/main.css
+++ b/src/assets/styles/main.css
@@ -16,4 +16,7 @@
     div:focus {
         @apply ring-inset;
     }
+    li:focus {
+        @apply ring-inset;
+    }
 }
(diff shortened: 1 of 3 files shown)


TishamIslam (changes requested):
image
The focus styling for this is a little off, the solution would be to go to src/assets/styles/main.css and where it has the styling for div, include li for it (i.e. `div:focus, li:focus {. . .}` )

Aside from that, nice catch!

DaceyTom2:
Ahh ty @TishamIslam should be good to go

TishamIslam:
Looks great!

TishamIslam (approved):

PR #104, opened 2021-07-22
Semantic markup for lists instead of divs

DaceyTom2:
### Description
Both the chatbot conversation and email messages now use ul and li tags instead of divs for accessibility reasons

### DIFF
--- a/src/assets/styles/main.css
+++ b/src/assets/styles/main.css
@@ -13,7 +13,7 @@
         @apply focus-visible:ring-1;
         @apply focus-visible:ring-black;
     }
-    div, li {
+    div, ul, li {
         @apply focus:ring-inset;
     }
     input {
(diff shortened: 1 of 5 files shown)


TishamIslam (changes requested):
Seems like it has the same styling issue as last time, tabbing to the conversation window doesn't give it the focus state. You should pull and re-merge with main (otherwise you might have merge conflicts with the PR I just merged) and add ul to main.css at `div, li {...}` so it becomes `div, li, ul {...}`

shewood (commented):
I saw the same as Tisham, you can't tab into the message list anymore.

DaceyTom2:
Good catch, I thought we would be navigating through conversation messages by default but it looks like we hit return to navigate into the conversation messages

nibivi77 (approved):
Much better! Thank you for this one.

TishamIslam (approved):
Looks good, everything's in order.

PR #105, opened 2021-07-22
Vcon 170 bubble animation

nibivi77:
### Description
Added small animation for sending/receiving messages.

### Additional Notes
This is done with pure CSS. 
We can modify the speed and where the bubble appears from as we see fit.

### Known bug

- After sending messages, switch from job banks to virtual assistant chat and see the animation appear again. Any ideas to get around it?

### DIFF
--- a/src/assets/styles/main.css
+++ b/src/assets/styles/main.css
@@ -17,3 +17,43 @@
         @apply ring-inset;
     }
 }
+
+/* Chat bubble animations*/
+.moveUp {
+    animation: slideUp 0.167s cubic-bezier(.4,0,.2,1);
+}
+
+.userMessage{
+    animation: slideUpAndRight 0.167s cubic-bezier(.4,0,.2,1);
+}
+
+.botMessage {
+    animation: slideUpAndLeft 0.167s cubic-bezier(.4,0,.2,1);
+}
+
+@keyframes slideUp {
+    from {
+        transform: translateY(44px);
+    }
+    to {
+        transform: none;
+    }
+}
+
+@keyframes slideUpAndRight {
+    from {
+        transform: translateX(-100px) translateY(44px);
+    }
+    to {
+        transform: none;
+    }
+}
+
+@keyframes slideUpAndLeft {
+    from {
+        transform: translateX(100px) translateY(44px);
+    }
+    to {
+        transform: none;
+    }
+}
--- a/src/assets/styles/main.css
+++ b/src/assets/styles/main.css
@@ -17,30 +17,3 @@
         @apply ring-inset;
     }
 }
-
-/* Chat bubble animations*/
-.userMessage{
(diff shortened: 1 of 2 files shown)


neilmispelaar:
Quick question - is the idea that these styles would be used anywhere else outside of the chat component? 

The reason I ask is because, another pattern to explore, is to put styles that are only used by a component right within the component. 

Note: this is just me being very opinionatd - and there is absolutely nothing wrong with putting styles in main.css 

Documentation for Single file components 
https://v3.vuejs.org/guide/single-file-component.html#introduction

image

neilmispelaar:
This is amazing! Can't say how awesome this is to see in action. 

And, it's so fun to see the bubbles come in. Well done! 

My feedback - but this is purely subjective and I'm sure @ThomasFHoy will have lots more to say:

- The bot bubbles should just come straight up (not up and travel to the left) 
- The user bubbles should also come up more from the left (it feels like they have too long a distance to travel to the right)

Otherwise wow this is amazing and so exciting to see! Yay. 

https://user-images.githubusercontent.com/9325038/126711883-f072f24a-7f88-4d0f-ab75-e5b5cd11ebe9.mov

DaceyTom2 (approved):
Well done Ben, it really gives the chat some life. While I do agree with Neil that the animations are better suited in the local style since there isn't much reuse value, it isn't a big stink.

I'll approve it's functionality but still check with Hoy that it matches his vision

nibivi77:
> Quick question - is the idea that these styles would be used anywhere else outside of the chat component?
> 
> The reason I ask is because, another pattern to explore, is to put styles that are only used by a component right within the component.
> 
> Note: this is just me being very opinionatd - and there is absolutely nothing wrong with putting styles in main.css
> 
> Documentation for Single file components
> https://v3.vuejs.org/guide/single-file-component.html#introduction
> 
> image
> 
> Note: using tailwind `@import` statements in the <style> section can be problematic (i.e. it doesn't work great)

You are right with this one. I was still in the mentality of Next.js and had forgotten about the scoped attribute.I will move it to the component.

nibivi77:
> This is amazing! Can't say how awesome this is to see in action.
> 
> And, it's so fun to see the bubbles come in. Well done!
> 
> My feedback - but this is purely subjective and I'm sure @ThomasFHoy will have lots more to say:
> 
> * The bot bubbles should just come straight up (not up and travel to the left)
> * The user bubbles should also come up more from the left (it feels like they have too long a distance to travel to the right)
> 
> Otherwise wow this is amazing and so exciting to see! Yay.
> 
>  Screen.Recording.2021-07-22.at.5.26.55.PM.mov

I will see what I can do with the bubbles placement. :)

neilmispelaar:
> it isn't a big stink

This is the greatest line ever

PR #106, opened 2021-07-22
Change lang link

shewood:
### Description
made it so that the screen reader reads the language that it would be changed into if you click the change language link.
translated the skip nav link.


neilmispelaar (approved):
Great work! 

Looks perfect to me.

PR #107, opened 2021-07-22
Vcon 172 inbox animation [Closing only]

nibivi77:
### Description
Built the animation to close the inbox on mobile.

### Additional Notes

 - I am a little bit stumped with the open section. Either I make it in a separate PR, or with this one when I have done it.


shewood (commented):
sad news is that the UI doesn't draw until you click one of the inbox items. 
image

DaceyTom2:
The animation is very clean. Might be a dumb questions but have you tried rendering the window off screen and transitioning it in after?

nibivi77:
Turns out the animation will change. Hoy suggested that the full screen should slide from left to right. I will see what I can do. Closing the PR for now.

PR #109, opened 2021-07-23
Get rid of useless logic now that the text input has no focus styles 

TishamIslam:
### Description
Pretty much what the title says, now that we don't use focus styles for the text inputs, we can get rid of the logic for checking if there's a click on them and if it was focused on by tabbing. Functionally changes nothing, just cleans up the code a little.


DaceyTom2 (approved):
Good job cleaning up! There are no obvious changes to functionality

PR #110, opened 2021-07-23
Bubble chat animation partial fix

nibivi77:
### Description
After sending messages to the chatbot, swapping from the chatbot to job banks reactivates all animations.
this PR makes it so that only the last message sent/received has the animation.

### Additional Notes
 - I am still working on removing all animation once it has been sent/recieved but for now, this is better.

### DIFF
--- a/src/components/atoms/ConversationMessage.vue
+++ b/src/components/atoms/ConversationMessage.vue
@@ -3,8 +3,13 @@
     <p
       :class="[
         !isUser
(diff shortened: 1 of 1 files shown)


shewood (approved):
This looks really good. I assume that you are still working on that last bubble so that it doesn't animate when you change the messages.

nibivi77:
> This looks really good. I assume that you are still working on that last bubble so that it doesn't animate when you change the messages.

Yes I am. For now this is better than all of the bubbles moving up.

PR #113, opened 2021-07-26
Fixing overflow-x on mobile.

nibivi77:
### Description
This PR fixes the overflow-x that the mobile version had.

### Additional Notes

- test in multiple browsers and desktop/mobile version as well.
- I also threw in the z-10 index for the header in the chatbot so that no messages would overlap on top of the header.


DaceyTom2 (approved):
Looks really good. Tried in FF and Chrome

PR #114, opened 2021-07-26
updated the placeholder text colour

shewood:
### Description
updated the placeholder text colour and also the font and input text colour. 

### Additional Notes
I did not also do the search bar as it is to be removed (in another task)

### DIFF
--- a/src/components/atoms/TextInput.vue
+++ b/src/components/atoms/TextInput.vue
@@ -6,7 +6,9 @@
     class="
       w-full
       border-t border-b border-gray-200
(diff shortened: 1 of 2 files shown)


DaceyTom2 (approved):
Looks good, I'm wondering if "search canada.ca" text colors should be updated as well?

shewood:
> Looks good, I'm wondering if "search canada.ca" text colors should be updated as well?

The search bar is going to be removed as per T.H. there is another task for it.

PR #115, opened 2021-07-26
removed the search bar from the header

shewood:
### Description
Removed the search bar from the header. 
(I left in the translation in case we need to add it back but can remove them too if wanted)


DaceyTom2 (approved):
So much empty!

PR #116, opened 2021-07-27
make inbox items scroll again

TishamIslam:
### Description
image
While doing stuff to do with text resizing, I noticed that the French version of the site at 200% zoom made the inbox items clip like in the picture above. So, I went through some old PRs, copied the formatting that had the scrollbar working in similar cases, and then removed as much of it as possible while keeping functionality intact. The CSS classes might be able to be reduced even further, but this seemed like enough. 

TL;DR scrolling fixed for inbox items. This won't be noticed unless you change some settings in-browser.

What it looks like in French mode + 200% zoom now:
image

### Additional notes
Also changed it so that the inbox list would overflow-scroll automatically starting from small screen sizes (since that's when we switch to the desktop view now).
Also removed a redundant class for making the inbox items get a black focus border (which they have by default now).

### DIFF
--- a/src/components/molecules/Inbox.vue
+++ b/src/components/molecules/Inbox.vue
@@ -4,14 +4,14 @@
     <div class="px-3 pt-6 pb-3">
       <div class="bg-gray-infolt h-10 w-2/3 animate-pulse"></div>
     </div>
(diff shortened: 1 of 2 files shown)


nibivi77 (approved):
Looks good to me! :)

PR #117, opened 2021-07-27
Translating emails / Job bank message

DaceyTom2:
### Description
Email is now translated from service. This is done through use of locale. Various dates are using locale to translate as well

### Additional Notes
[VCON-252 Translate Virtual Assistant to Assistant virtuel](https://decdvirtualconcierge.atlassian.net/browse/VCON-252?atlOrigin=eyJpIjoiMDhiY2I1ZmVjZjM3NGE1N2E4ZWUzZjY5ZDJmM2ZmYmIiLCJwIjoiaiJ9) is included in this change

### Test
1. View page in English and make sure all text corresponds
2. View page in French and make sure all text corresponds
3. Search through Inbox
4. *Chatbot messages aren't required to be translated as of yet as it exists outside of the VC front end


TishamIslam:
Looks great! Everything seems to be in order, but I'm not 100% sure if the message is supposed to be become unread when you switch languages? I guess it makes sense since it seems you have to reload the whole thing and that most people probably won't encounter that issue to begin with, but just wanted to make sure if that was taken into consideration.

nibivi77 (approved):
This is good. I was just wondering if we could have consistency in the inbox last read date.

In English its all cap with no periods at the end of the abreviation.
image


In French there is no cap with an period.
image

DaceyTom2:
@TishamIslam I talked with Farwa about how the app should act. The best I could do is reloading all state but it has an unfortunate side effect. In a production environment this would be hooked up to a back end and we would have a way to persist user data. Unfortunately that is out of scope for this prototype.

@nibivi77 I was under the impression English used caps and French was lower with a period. If there is a better way please let me know.

PR #118, opened 2021-07-28
added jest and vue test utils

shewood:
### Description
Added Jest, Vue test utils and babel along with config files. 
Added one test. 

### Additional Notes
you will need to run npm i before using. 
It is not yet hooked up to husky.


__tests__/banner.spec.js
```diff
@@ -0,0 +1,18 @@
+import { mount } from '@vue/test-utils'
```
DaceyTom2, line 1: It would be helpful if the test directory followed a similar patter to src so we can navigate easily. ie /\_\_tests\_\_/components/atoms/Banner.spec.js

DaceyTom2 (changes requested):
Nice job! One small change for structure

__tests__/banner.spec.js
shewood: Good idea.

DaceyTom2 (approved):

PR #119, opened 2021-07-28
- unnecessary html/css - lang="ts" + one translation

TishamIslam:
### Description
Removed/moved/modified html/css content such that there were no more functional redundancies. Might've missed a few less obvious ones.

Removed the `lang="ts"` in some of the script tags, as we're not using TypeScript. 

Added a translation for the SkipNav's aria-label. 

### Additional Notes
I removed the script section entirely for the DateModified component, as it wasn't actually doing anything.

I was considering removing the setup() methods in some of the script files where they weren't used for anything, but wasn't sure if it was worth it in case those elements needed one in the future. Let me know if you think that removing them is actually a good idea (applies to stuff like FeedbackButton, Menu, etc. which might need them later down the line)

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -2,7 +2,7 @@
   <router-view />
 </template>
 
-<script lang="ts">
+<script>
 export default {
   name: "App",
   components: {},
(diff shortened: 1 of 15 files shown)


DaceyTom2 (approved):
Great cleanup! Still functions well. I'd say removing setup when not used is fine. Easy to add them back if needed

PR #120, opened 2021-07-28
use disable instead of tabindex

TishamIslam:
### Description
Instead of disabling the send button by setting tabindex to 0, I used the disable attribute.


nibivi77 (approved):
I like this. This is cleaner. Good Job Tisham!

PR #121, opened 2021-07-29
Added jest-quick-fix to import components.

nibivi77:
### Description

Added quick line to parse Vue component files.

Split in half the inbox open and close animation time.


shewood (approved):
looks good. Thanks for putting that back in.

PR #123, opened 2021-07-29
updated the test for banner

shewood:
### Description
Updated the banner tests with additional props and imports.


nibivi77 (approved):
Good :+1:

PR #127, opened 2021-07-30
Vcon 167 mound vue and test for h1

DaceyTom2:
### Description
Add a simple unit test that mounts vue and tests that the app has an 'h1'

### Additional Notes
Added mocking for chatMessages module
Added mocking for emails module
Added plugin for i18n mount


shewood (approved):
looks good to me

PR #131, opened 2021-08-02
Removed TheSkipNav

DaceyTom2:
### Description
Removed unused code


TishamIslam:
This is needed as per [WCAG 2.4.1](https://www.w3.org/WAI/WCAG21/quickref/#bypass-blocks), you can only access it through focusing on it by tabbing, after which it becomes visible.

DaceyTom2:
You are going to have to show me where this is used. My mind is blown

neilmispelaar:
Screen Shot 2021-08-02 at 11 28 37 AM

Screen Shot 2021-08-02 at 11 28 32 AM

neilmispelaar:
I can't recommend the removal of SkipNav links but I can add that the focus styles weren't added in (I think the boilerplate just had the hidden links)

TishamIslam:
I can't send the screen recording here but I'll it send it to you on slack as well, Neil.

DaceyTom2:
I miss-understood how this was being used in the code

PR #132, opened 2021-08-02
Added message card tests

DaceyTom2:
### Description
MessageCard test portion of VCON-280 using shallow mount and i18n plugin
Please let me know if there are other opportunities to create more robust tests

### Additional Notes
Props added and checked to see if rendered anywhere on the page


nibivi77 (approved):
Not much else top test here! Good job!

PR #133, opened 2021-08-02
Added inbox item tests

DaceyTom2:
### Description
InboxItem test portion of VCON-280 using shallow mount and i18n plugin
Please let me know if there are other opportunities to create more robust tests

### Additional Notes
Props added and checked to see if rendered anywhere on the page
Click inbox item tests emits
Keyup enter tests emits

### DIFF
--- a/__tests__/components/molecules/InboxItem.spec.js
+++ b/__tests__/components/molecules/InboxItem.spec.js
@@ -1,0 +1,52 @@
+import { shallowMount } from "@vue/test-utils";
+import { i18n } from "../../../i18n";
+import InboxItem from "../../../src/components/molecules/InboxItem";
+import ReadNotification from "../../../src/components/atoms/ReadNotification";
+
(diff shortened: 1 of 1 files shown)


shewood (approved):
Looks good to me.

nibivi77 (approved):
This is great! You could also check that when you click on an inbox item, the ReadNotification component is gone and it is replaced with a date. But this is good!

nibivi77:
Another thing you can test is are the classes applied when clicked or not. This would ensure visual differences.

DaceyTom2:
@nibivi77 the date change is done within the ReadComponent. I rather keep it shallowMounted but that test can be done within the ReadComponent tests

DaceyTom2:
I'm a bit sceptic to add tests to see if CSS is applied. The logic is quite minimal and it is more overhead to update tests for design changes. The CSS also doesn't break the functionality of the app itself

PR #134, opened 2021-08-03
fix skip nav "pushing" page downwards

TishamIslam:
## No Jira issue, mentioned by Neil 
image

### Description
Changed the skip nav so it'd have absolute positioning, therefore it no longer "pushes" the page as it doesn't appear on top of the header. 

#### Then (Note the whitespace): 
image

#### Now:
image

#### Canada.ca comparison:
image

### Additional Notes
Took the liberty of fixing up the inbox's focus style, part of the outline wasn't visible so I used an inset ring instead.
Also note that due to the absolute positioning, at certain screen sizes the skip nav will go over the government logo, BUT canada.ca also has that behaviour, so it's probably irrelevant. 
image

### DIFF
--- a/src/components/molecules/Inbox.vue
+++ b/src/components/molecules/Inbox.vue
@@ -14,7 +14,7 @@
   <div v-else class="w-full sm:h-vh-3/5 xl:h-vh-2/3 flex flex-col">
     <h1
       id="inbox-header"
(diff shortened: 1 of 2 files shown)


nibivi77 (approved):
Nice job!

PR #135, opened 2021-08-03
fixed inbox item states

TishamIslam:
### Description
Basically had these things in mind:
- selected state stays when clicking away
- selected state appears on the default selected item
- hover and focus doesn't change an inbox item that's already selected

### Hoy Seal of Approval 
image


nibivi77 (approved):
Nice good catch!

PR #136, opened 2021-08-03
properly hide drawer items when not in view

TishamIslam:
### Description
Focus wasn't visible due to drawer elements being tabbable when the drawer wasn't in view, so I used the hidden class on the drawers when they weren't in view so that way they weren't tabbable. This should also prevent screenreaders from reading the drawer item contents. 

### Additional Notes
Not sure if this actually relates to VCON-142 (WCAG 2.4.7 Focus visible) or if there's a different issue for elements that aren't in view that can still receive focus. Regardless, issue resolved.
I tried to do the `isMobileDrawerOpen ? 'flex' : 'hidden'` thing in the MessagesContainer as opposed to the individual ConversationWindow and MessageList components, but it also removed the sliding animation entirely. I imagine there's a better way to do this using different animations and such, but I know very little about animation with css/tailwind. Feel free to try out something like that yourself.

### DIFF
--- a/src/components/molecules/MessageList.vue
+++ b/src/components/molecules/MessageList.vue
@@ -1,5 +1,10 @@
 <template>
-  <div class="w-full h-full flex flex-col">
+  <div
+    :class="[
+      'w-full h-full sm:flex flex-col',
+      isMobileDrawerOpen ? 'flex' : 'hidden',
+    ]"
+  >
     <message-header
       backIcon="Back"
       :imageName="mailObject.senderIcon"
@@ -43,6 +48,10 @@ export default {
         selectedInboxItemId
       )
     );
+
+    const isMobileDrawerOpen = computed(
+      () => store.getters["inbox/isMobileDrawerOpen"]
+    );
     //creates a timestamp with js's default date methods
     const createTimestamp = (dateString) => {
       return new Date(dateString).toLocaleDateString(i18n.global.locale.value, {
@@ -56,6 +65,7 @@ export default {
     return {
       mailObject,
       createTimestamp,
+      isMobileDrawerOpen,
       icons,
     };
   },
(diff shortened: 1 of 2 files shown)


nibivi77 (commented):
I assume this will be for screen readers after Pat comes back from vacation? Did he say this would be better? I think it would but just to get confirmation you know?

Instead of using a conditional class `hidden` and to reduce code repetition, we could use `v-show` like this:

```  
<div
    class='w-full h-full sm:flex flex-col'
    v-show="isMobileDrawerOpen"
  >
```

Also, the first inbox item is not selected on the first page load. Is this on my end?

 That's just a suggestion so you don't have to do it. Apart from that This looks good.

neilmispelaar:
Haven't had a chance to do a deep dive, but wanted to share this link to show (pun intended) the difference between `v-if` and `v-show`

https://www.reddit.com/r/vuejs/comments/otdn7s/vue_academy_3_vif_vs_vshow/

nibivi77:
Great link Neil! Thank you.

this is what we need to extract from the link.

image

TishamIslam:
> I assume this will be for screen readers after Pat comes back from vacation? Did he say this would be better? I think it would but just to get confirmation you know?

Actually, this is just a bit I took out of that PR. This bit felt a lot more unambiguous in terms of "what's the right thing to do" so I made it into a separate PR while we wait for Patrick to come back for his input on the other PR. 

> image

I'll test on Voiceover with v-show, I don't remember off the top of my head if `display: none` hides the text from it.

TishamIslam:
Voiceover doesn't read it, so that's good to know as we can use it elsewhere if need be.

nibivi77 (approved):
Looks good to me!

The only thing i'll say is that if our app becomes bigger, hiding everything in the back when the menu is open might not be the best approach. 

For now this is great.

PR #137, opened 2021-08-03
Fixed conversation window not displaying on page load

TishamIslam:
### Description
Merged my latest PR (#136) without checking something, my bad. I switched out v-show with the previous method of adding classes conditionally, as the breakpoint class `sm:flex` will come into effect on the desktop view and display the window as normal.

### DIFF
--- a/src/components/organisms/ConversationWindow.vue
+++ b/src/components/organisms/ConversationWindow.vue
@@ -1,7 +1,9 @@
 <template>
   <div
-    class="w-full h-full flex flex-col text-gray-dark sm:relative"
-    v-show="isMobileDrawerOpen"
+    :class="[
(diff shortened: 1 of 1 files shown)


nibivi77 (approved):
Ah yes that is why we did not use v-show. Sorry about that! 

Do we need to switch out the `MessageList` v-show as well? Other then that, thank you!

TishamIslam:
> Ah yes that is why we did not use v-show. Sorry about that!

Really it was my mistake, I forgot to address that part of your comment on the PR. 

> Do we need to switch out the `MessageList` v-show as well? Other then that, thank you!

I don't think that we need to switch out the MessageList's v-show unless we wanted to be consistent. As the ConversationWindow is the only component that needs this functionality of showing on page load, I think it makes sense to keep the difference.

PR #138, opened 2021-08-04
Test messageheader

shewood:
### Description
Added a test for the Message Header 

### Additional Notes
there might be a better way to use the store instead of the before all.


TishamIslam (approved):
Looks good!

PR #139, opened 2021-08-04
added test for feedback button

TishamIslam:
### Description
Added a test for the feedback button. I don't know if it's necessary to also check if the colorClass property is added to the button properly.


nibivi77 (approved):
Yep this works!

PR #140, opened 2021-08-05
VCON-Added tests to inbox store

DaceyTom2:
### Description
Unit tests added to Vuex state for inbox

### Additional Notes
Calls to chatMessages and emails have been mocked
Inbox getters are a bit wonky as I had to use getters from chatMessages for a decent amount of them so they follow what is in the mock files

### DIFF
--- a/__tests__/store/modules/inbox.spec.js
+++ b/__tests__/store/modules/inbox.spec.js
@@ -89,6 +89,7 @@ describe("inbox actions", () => {
   it("selectDefaultInboxItem", () => {
     store.dispatch("inbox/selectDefaultInboxItem", 5);
 
+    //Test argument being passed to mutation correctly
     const updateSelectedInboxItemIdArgs =
       mockedMutations.updateSelectedInboxItemId.mock.calls[0][1];
     expect(updateSelectedInboxItemIdArgs).toBe(5);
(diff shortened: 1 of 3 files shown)


__tests__/store/modules/inbox.spec.js
```diff
@@ -0,0 +1,134 @@
...
+
+describe("inbox getters", () => {
+  it("getInboxItems", () => {
+    const store = createStore({
+      modules: {
+        inbox,
+        emails,
+        chatMessages,
+      },
+    });
```
TishamIslam, line 16: Seems like you could put this into a `beforeEach()` as you did for the second set of tests, since 4/5 tests in this section need this exact bit of code. The `isMobileDrawerOpen` test could then be done as a separate test.

DaceyTom2: I was on the fence as you get more flexibility if you recreate it and some cases we could take advantage of it. However, since there is a preference I changed it. I made it a beforeAll so it is only created once. isMobileDrawerOpen doesn't use the store so it should be fine. The reason I had to use the store is because there are dependencies on other modules that break without it.

TishamIslam (approved):
Looks great, I don't understand the lines like `const updateSelectedInboxItemIdArgs =
      mockedMutations.updateSelectedInboxItemId.mock.calls[0][1]` but aside from that everything else makes sense and I can see what they're testing.

DaceyTom2:
I added a comment to help explain. It checks that the argument passed to the mutator is correct

PR #141, opened 2021-08-05
removed ReadNotification's id values entirely

TishamIslam:
### Description
Removed the id values in the ReadNotification component, as it raised an automatic issue in the Axe devtools for duplicate ids. 
Reproduction: Go to the [main site](https://virtual-assistant.toadfor.ca/en), click on the inbox item for Job Bank, and you should see the error in the Axe scan. 

### Additional Notes
Dunno if there's some functionality attached to this id attribute, I didn't see any.

### DIFF
--- a/__tests__/components/atoms/ReadNotification.spec.js
+++ b/__tests__/components/atoms/ReadNotification.spec.js
@@ -2,28 +2,19 @@ import { mount } from "@vue/test-utils";
 import ReadNotification from "../../../src/components/atoms/ReadNotification";
 
 describe("ReadNotification component", () => {
+  const dayReadProp = "Monday";
   it("ensures that the read notification is displayed", () => {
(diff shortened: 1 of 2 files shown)


DaceyTom2 (commented):
I believe these are used in tests to see which div is rendered

TishamIslam:
The test shouldn't rely on the id values anymore.

DaceyTom2 (changes requested):
Can the test be changed to check if the prop dayRead is displayed on the page instead? Would make it more robust and less abstract on what the purpose of the component is

TishamIslam:
Removed the third test since it seemed to more or less do the same thing as the first test, but by using a data function instead of a prop.

DaceyTom2 (approved):
Nice! This will be a solid test to use an example

PR #142, opened 2021-08-06
Vcon 168 unit test emails vuex state

DaceyTom2:
### Description
Added tests for vuex email module

### Additional Notes
AddEmail test was skipped as it isn't implemented in the code
State was crafted for getter tests
Email service was mocked to avoid integration dependencies


nibivi77 (approved):
Beautifully done Tom. I am learning through this PR. Superb!

PR #143, opened 2021-08-06
Vcon 168 unit test chatmessages vuex state

DaceyTom2:
### Description
Added tests for vuex chatMessages module

### Additional Notes
State was crafted for getter tests
chatMessages service was mocked to avoid integration dependencies


nibivi77 (approved):
This is great again! (Made a small change to one of the tests because we changed the ordering of the returned value in one of the mutators.)

I also tried merging with main but I forgot to sign it, so I tried signing it again.

PR #144, opened 2021-08-06
un-reversed / re-reversed conversation window

TishamIslam:
### Description
Re-reversed/un-reversed the conversation window's chat messages so they should now be in order when read out with a screen reader/not using css. 

### Additional Notes 
Thanks to [stack overflow](https://stackoverflow.com/questions/36130760/use-justify-content-flex-end-and-to-have-vertical-scrollbar) for always being there, even when I search up words that don't make proper sentences ♥. 
Only checked if the conversation started from the bottom and if the scrollbar didn't disappear, not sure if there's any other issues to check for/browser compatibility.

### DIFF
--- a/src/components/organisms/ConversationWindow.vue
+++ b/src/components/organisms/ConversationWindow.vue
@@ -16,9 +16,9 @@
       <ul
         class="
           w-full
-          space-y-2 space-y-reverse
+          space-y-2
           overflow-auto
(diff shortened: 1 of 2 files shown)


nibivi77 (approved):
Seems to work just fine on Chrome and Firefox. Good job Tisham ! :)

PR #145, opened 2021-08-09
Get quick replies/suggested actions from bot into chatoption buttons

TishamIslam:
### Description
Got the suggestedActions thing from the bot, which allows us to use the quick replies given to us by the bot. 

### Additional Notes
At the moment I have it set to default to "Yes" or "No"/"Oui" ou "Non" as the default quick replies if none are sent from the bot.
For testing, at the moment it seems like only the prompt to rate the bot has suggestedActions. So, to test it efficiently, hit the No button. If the prompt gets sent in the wrong order (i.e. it isn't the last message) press No again.

#### Nothing sent by bot so it defaults to Yes/No
image

#### Same thing but en français
image

#### Success!
image

### DIFF
--- a/src/components/atoms/TextInput.vue
+++ b/src/components/atoms/TextInput.vue
@@ -22,7 +22,7 @@
       <chat-option-button
         v-for="(option, index) in buttonOptions"
         :key="index"
-        :text="option"
+        :text="option?.value ?? option"
         @send-button="sendBtnText"
       />
     </div>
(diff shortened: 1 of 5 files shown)


nibivi77 (commented):
this is great Tisham! I think it would be better to not show any suggested answers if none are provided. (hide the buttons)

What do you think?

TishamIslam:
Yup that makes sense, the user can enter yes or no manually anyways.

nibivi77:
> Yup that makes sense, the user can enter yes or no manually anyways.

Exactly, and "Yes" and "no" will most likely be there as suggested actions in the future.

nibivi77 (approved):
This looks good ! :)

PR #146, opened 2021-08-10
updated timestamps for message cards

TishamIslam:
### Description
Fixes 2 main things:
- mail thing (job bank) not working at all
- scrolling in the chat window on mobile

There's a few smaller things but those were already implemented before, just reverted some things back to their natural state.


nibivi77 (approved):
Thank you Tisham! This fixes a good amount of issues we found.

PR #149, opened 2021-08-11
fixed all the tests that broke

TishamIslam:
### Description
Fixes broken tests — They seem to all have broken due to certain changes in implementation... will try to figure out ways to make the tests less reliant on implementation details as well. Some broke partially or entirely due to not including the i18n plugin.

Quick run-down on why each test broke:
MessageCard: props weren't updated in tests
ChatMessages: p tag was changed to h4; test tried to find p tag unsuccessfully. also i18n was used for hidden text
ChatOptionButton: i18n was used for hidden text
ConversationWindow: mocked messages is supposed to be an array; wasn't, so it failed
ReadNotification: i18n was used for hidden text


shewood (approved):
Nice work!

PR #150, opened 2021-08-11
fixed link for back to projects

shewood:
### Description
Fixed the link for Back to projects so that it now goes to the alpha site projects page.

### DIFF
--- a/src/components/atoms/PhaseBanner.vue
+++ b/src/components/atoms/PhaseBanner.vue
@@ -29,7 +29,6 @@ export default {
   name: "PhaseBanner",
   props: {
     phase: String,
-    href: String,
     link: String,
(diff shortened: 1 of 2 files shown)


src/components/atoms/PhaseBanner.vue
```diff
@@ -29,6 +29,7 @@ export default {
   name: "PhaseBanner",
   props: {
     phase: String,
+    href: String,
```
TishamIslam, line 32: I don't think this is needed.

TishamIslam (changes requested):
Works as intended now, but you seem to have added an unused prop btw.

TishamIslam (approved):
Perfect!

PR #153, opened 2021-08-12
focus on main content on inbox item click

TishamIslam:
### Description
Focused on the main content when you select an inbox item. This focus is not visible when you click on an inbox item (since most of our focus styles use focus-visible)

### DIFF
--- a/src/components/molecules/Inbox.vue
+++ b/src/components/molecules/Inbox.vue
@@ -45,7 +45,12 @@ export default {
     const inboxItems = computed(() => store.getters["inbox/getInboxItems"]);
     const inboxLoaded = computed(() => store.getters["inbox/isLoaded"]);
     function selectInboxItem(index) {
-      store.dispatch("inbox/selectInboxItem", index);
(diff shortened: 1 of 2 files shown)


neilmispelaar (approved):
Excellent stuff! 

If I can recommend for future considerations: 

- chaining is cool (and I love it) but less readable 
- in vue it's sometimes necessary to wait for what's called "nextTick" which is an update cycle in vue. This gives vue a moment to do it's thing in the backend and then update the dom so that when you go to focus on a thing that thing has been correctly inserted in the dom. 

Here is a code sample from here https://michaelnthiessen.com/set-focus-on-input-vue
```
// Focus the component, but we have to wait
      // so that it will be showing first.
      this.nextTick(() => {
        this.focusInput();
      });
```

TishamIslam:
Lol I love chaining as well, if it were possible and if it wouldn't be a violation of basic human rights, I'd totally just chain everything into super verbose one-liners everywhere

neilmispelaar:
LOL

PR #154, opened 2021-08-12
Small cleanup + removed redundant roles

TishamIslam:
### Description
Cleaned up a bit of code and also removed the list/listitem roles that were redundant. 

### Additional Notes
Also updated PR template to use VA-XXX instead of VCON-XXX


shewood (approved):
Looks good

PR #156, opened 2021-08-12
Adding aria label to unread notification icon

karthikktamilmani-freshworks:

### DIFF
--- a/src/components/atoms/ReadNotification.vue
+++ b/src/components/atoms/ReadNotification.vue
@@ -25,7 +25,10 @@
     v-else
     class="flex place-items-center justify-center h-16 w-16 md:h-20 md:w-20"
   >
-    <div class="bg-alert-main rounded-full w-4 h-4" />
+    <i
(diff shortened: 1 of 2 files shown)


neilmispelaar (changes requested):
Hi Karthikk, 

#### ReadNotification.vue: 

The `<i>` tag has semantic meaning so I can't recommend it's use in this particular case. 

Would be much better to use a `<span>` or `<div>` element here. 

#### InboxItem

Will approve but would ask that maybe we can take a moment to explain what's going on exactly. 

Thank you! Looking great.

TishamIslam (changes requested):
Hi Karthikk, I believe what's meant for this task is to provide not just hidden text indicating that there's an unread message but to also provide visible text for sighted users that may not understand the significance of the blue icon. 

"Many users may have difficulties determining the meaning of this icon based on the colour alone." implies that the Accessibility team wants us to account for sighted users and not just users using screenreaders.

PR #157, opened 2021-08-12
changed heading levels and changed headers to spans where applicable

TishamIslam:
### Description
Heading level changes: 
- Inbox heading from h1 to h2 (I'm assuming we're going to bring the Service Canada Labs banner back at some point)
- Inbox sender name from h2 to h3
- Message card from h2 to h4 (as it is semantically under the sender name). Note that this may raise an Axe devtools issue 

Changed headers to spans for:
 - readnotification
 - chat message
 - message preview in the Inbox component

### Additional Notes
Axe devtools issues are raised for the spans as they have aria-labels attached to them — sr-only text may be more suitable

### DIFF
--- a/src/components/atoms/ConversationMessage.vue
+++ b/src/components/atoms/ConversationMessage.vue
@@ -8,8 +8,7 @@
     @keydown.prevent.down="nextElement($event)"
     @keydown.prevent.exact.tab="nextElement($event)"
   >
-    <!-- <span class="sr-only">{{ (isUser ? "You " : senderName) }} said:</span> -->
-    <h4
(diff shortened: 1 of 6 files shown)


shewood (commented):
Checkout the changes that are in PR #156 as it looks like you and Karthikk are changing the same files for the read notification.

shewood (changes requested):
I think we need the h1 for now as the other heading is hidden

shewood (approved):
Approved for now but when you changed the headings the classes changed for the styles and they will need to be fixed at some point.

PR #158, opened 2021-08-12
French translations 

TishamIslam:
### Description 
Mostly resolves hidden text, I searched through everything that had arialabels and aria-labelledby attributes. might have missed some visible text bc of that.


neilmispelaar (approved):
Everything looks good to me. 

I did not test locally due to time constraints but could not see anything offensive in the code

PR #159, opened 2021-08-12
Adding title for virtual assistant page

karthikktamilmani-freshworks:


neilmispelaar (approved):
This works!  Awesome sauce - with one note for the future below: 

While I was doing the review I noticed that the language toggle link is using an `a` anchor as opposed to a `<router-link>`

And so - when a language toggle occurs, the whole page reloads and that's why the page title changes in the `created` event handler. 

Ideally, when we switch the language toggle link to use a `router-link` - I don't think the `created` event will work to update the title because the component isn't being recreated like when the page is refreshed.

shewood:
That is a good comment about the router-link.

PR #160, opened 2021-08-12
VA-78 translated alt text for header and footer

shewood:
### Description
resolves the issue with the not very descriptive alt text for the canada logos in the header and footer. 
_Providing a text alternative that is not null (e.g., alt=spacer" or alt="image") for images that should be ignored by assistive technology"_
translated the text.

### DIFF
--- a/src/components/molecules/TheFooter.vue
+++ b/src/components/molecules/TheFooter.vue
@@ -18,7 +18,7 @@
             <img
               className="h-6 sm:h-8"
               src="../../assets/wmms-blk.svg"
-              alt="canada logo"
+              :alt="$t('canadaLogoFooter')"
             />
           </div>
         </div>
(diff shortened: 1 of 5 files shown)


neilmispelaar (dismissed):
Awesome stuff! Thanks for the quick turn around time. 

Two small changes requested with splash page: 

For the splash page the alt text for the bilingual Government of Canada word mark should be set to: 

"Government of Canada / Gouvernement du Canada"

I'm not exactly sure what it's set to now as it's pulling from a translation file and I can only see the `$t(key)` in the code. 

The reason it should be set to bilingual text is because there is no language selection on the splash page (the user can't toggle the splash page) so we just set the alt text to bilingual by default. 

Same would apply for the Canada word mark towards the bottom of the content square. That should have the following text:  

Symbol of the Government of Canada / Symbole du gouvernement du Canada

neilmispelaar (changes requested):
See comment about splash page changes. 

Thank you!

shewood:
I think the page pulls the language of the browser and would then use the appropriate translation but I just put them both in the alt text as suggested.

neilmispelaar (approved):

PR #161, opened 2021-08-13
VA-82 Added white outline around skip nav button

shewood:
### Description
I added a solid outline in white around the button. It's not exactly changing the black outline but it looks better and is more noticeable. Also moved it down on larger screens so that it hopefully will not be over the text. Small screens it covers some but still is noticeable.

### DIFF
--- a/src/components/molecules/TheSkipNav.vue
+++ b/src/components/molecules/TheSkipNav.vue
@@ -16,9 +16,11 @@
           bg-blue-deep
           text-white
           mt-2
+          sm:mt-14
           p-2
           rounded
           block
+          outline-white
         "
         @click="toMain"
       >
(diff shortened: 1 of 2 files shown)


neilmispelaar (approved):
Approved!

My only feedback is that if this could be done with [utility classes](https://tailwindcss.com/docs/border-style) than that would have been my preference as opposed to expanding the custom base of classes. 

That being said - both achieve their objectives which is to add the outline around the button so excellent stuff. 

I think also as soon as next week the whole top notice / skip navs / feedback bits are getting redone so don't worry about changing anything here.

neilmispelaar:
Thank you!

shewood:
The utility class is a dotted line, I made it solid for this occasion.

PR #164, opened 2021-08-13
VA-72 Add aria-live to message list

neilmispelaar:
### Description

Screen reader users are not notified when they send a message, and they are not notified about any of the new messages they received in the conversation section. They will need to navigate back and read all the messages in order to understand the conversation. Use aria-live to provide users with notifications. The following page contains guidelines about how to create accessible chat chatbot for all users: https://a11y-guidelines.orange.com/en/web/components-examples/chatbot/ 

Added `aria-live` to the ul message list 

### Additional Notes

Nope

### DIFF
--- a/src/components/atoms/ConversationMessage.vue
+++ b/src/components/atoms/ConversationMessage.vue
@@ -7,6 +7,7 @@
     @keydown.prevent.shift.tab="prevElement($event)"
     @keydown.prevent.down="nextElement($event)"
     @keydown.prevent.exact.tab="nextElement($event)"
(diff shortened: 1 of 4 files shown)


neilmispelaar:
Sheila's initial tests are showing a screenreader as NOT reading out messages - investigating

neilmispelaar:
Screen Shot 2021-08-13 at 10 00 16 AM

TishamIslam (approved):
Seems to work. Though it appears that my sent messages are read out as well.

